PR #1, opened 2018-05-24
add publish docs script

jckr:
I'm adding a script for convenience. type npm run / yarn + update-docs to build the ocular website then push the output to the gh-pages branch. 

note that for the script to work, you should have installed the node_modules in the website folder to begin with. That script won't do that step each time, so it must be done at least once.


georgios-uber:
Going to use a different script for this.

PR #4, opened 2018-06-07
Implement editable lineString layer

matthrice:
Followed the editablePolygonsLayer very closely to create a layer which represents geoJson lineStrings and allows selecting/editing the points.


georgios-uber:
> Forgot to run prettier? There are files without correct code style:
> src/lib/layers/editable-lines-layer.js

Please run prettier and update.

PR #25, opened 2018-07-06
Upgrade deck.gl to 6.0.0-beta.1

supersonicclay:


.gitignore
```diff
@@ -6,6 +6,7 @@ dist-es6/*
 node_modules/
 coverage/
 .nyc_output/
+.npmrc
```
georgios-uber: Why?

examples/deck/package.json
```diff
@@ -4,8 +4,7 @@
     "start-local": "webpack-dev-server --env.local --progress --hot --port 3000 --open"
   },
   "dependencies": {
-    "@deck.gl/experimental-layers": "5.X.X",
-    "deck.gl": "5.X.X",
+    "deck.gl": "6.0.0-beta.1",
```
georgios-uber: You should import the experimental layers. Then provide an option to use directed+shadowed for lines.

georgios-uber: #26

examples/webpack.config.local.js
```diff
@@ -32,10 +32,6 @@ function makeLocalDevConfig(EXAMPLE_DIR = LIB_DIR) {
         'nebula.gl': SRC_DIR,
 
         'deck.gl': resolve(LIB_DIR, './node_modules/deck.gl'),
-        '@deck.gl/experimental-layers': resolve(
```
georgios-uber: We want them!

src/lib/layers/segments-layer.js
```diff
@@ -1,99 +0,0 @@
-// @flow
-import { COORDINATE_SYSTEM } from 'deck.gl';
-import { PathMarkerLayer } from '@deck.gl/experimental-layers';
```
georgios-uber: This needs to be usable with #26

.gitignore
supersonicclay: This was the way I found to have npm use the public registry for this project while continuing to use internal npm for other projects. I place an `.npmrc` file where there's a `project.json` with the following contents:

```
registry=https://registry.yarnpkg.com
```

PR #30, opened 2018-07-11
add editType 'finishMovePosition'

matthrice:
Add new `editType = 'finishMovePosition'` to distinguish when a drag has finished and expose it in `onEdit` callback

### DIFF
--- a/docs/api-reference/layers/editable-geojson-layer.md
+++ b/docs/api-reference/layers/editable-geojson-layer.md
@@ -115,7 +115,7 @@ The `onEdit` event is the core event provided by this layer and must be handled 
 
   * `addFeature`: A new feature was added. Its index is reflected in `featureIndex`
 
-  * `finishMovePosition`: A position finished moving (indicated by `mouseup`).
+  * `finishMovePosition`: A position finished moving (indicated by `pointerup`).
 
 * `featureIndex` (Number): The index of the edited/added feature.
 
--- a/docs/api-reference/layers/editable-geojson-layer.md
+++ b/docs/api-reference/layers/editable-geojson-layer.md
@@ -113,7 +113,9 @@ The `onEdit` event is the core event provided by this layer and must be handled 
 
   * `removePosition`: A position was removed.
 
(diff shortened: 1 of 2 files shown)


src/lib/layers/editable-geojson-layer.js
```diff
@@ -383,6 +407,22 @@ export default class EditableGeoJsonLayer extends EditableLayer {
...
+  handleFinishMovePosition(featureIndex: number, positionIndexes: number, groundCoords: number[]) {
+    const updatedData = this.state.editableFeatureCollection
+      .replacePosition(featureIndex, positionIndexes, groundCoords)
+      .getObject();
+
+    this.props.onEdit({
+      updatedData,
+      updatedMode: this.props.mode,
+      updatedSelectedFeatureIndex: featureIndex,
+      editType: 'finishMovePosition',
```
georgios-uber: Please move all these in an Object that is exported.

matthrice: All of the `handle` functions?

docs/api-reference/layers/editable-geojson-layer.md
```diff
@@ -113,7 +113,9 @@ The `onEdit` event is the core event provided by this layer and must be handled
 
   * `removePosition`: A position was removed.
 
-  * `addFeature`: A new feature was added. Its index is reflected in `featureIndex`.
+  * `addFeature`: A new feature was added. Its index is reflected in `featureIndex`
+
+  * `finishMovePosition`: A position finished moving (indicated by `mouseup`).
```
supersonicclay: bit of a nit, but it is actually triggered by `pointerup`, not `mouseup`.

src/lib/layers/editable-geojson-layer.js
georgios-uber: @supersonicclay @matthrice Yes please do it for all of them.

georgios-uber:
@matthrice You can merge this one, don't worry about the constant.

PR #34, opened 2018-07-18
Allow panning map when dragging polygon fill

supersonicclay:
Previous behavior:
Dragging an edit handle suppresses panning map but instead moves the point.
Dragging a line or fill suppresses panning map but does nothing else.

New behavior:
Dragging an edit handle suppresses panning map but instead moves the point.
Dragging a line or fill doesn't suppress panning map. In the future, this may move the feature (perhaps only when also holding down Ctrl or Cmd)


georgios-uber:
Can we click on a line to edit?

I want to have that as an option or addition to handles.

supersonicclay:
The demo shows how clicking on a feature selects it. Is that what you mean? Or are you talking about adding a point by clicking anywhere along the line?

georgios-uber:
Yes.

supersonicclay:
If you're talking about adding a point by clicking anywhere along the line, I think we should add that as an issue and tackle in the future.

georgios-uber:
I want to have the option to either show handles or be able to click anywhere.

PR #36, opened 2018-07-23
Support IconLayer for editing handles

matthrice:
Saw Clay's `TODO` in one of the files and thought this would be a really useful addition.

### Icon Handles

Support the use of Deck.gl's `IconLayer` to render the editing handles as a provided icon. Adds the following properties to `EditableGeoJsonLayer`:

* `useIconsForHandles`
* `iconAtlas`
* `iconMapping`
* `getIcon`
* `getIconSize`

All of which are documented now. These can definitely be added to/removed from, but these are the ones I thought added value without adding too many new properties. Take a look at [IconLayer](https://github.com/uber/deck.gl/blob/master/docs/layers/icon-layer.md) to see if there are other props worth adding. 

The example is also updated with an icon toggle. Renders 8-balls and oranges as handles.

### DIFF
--- a/examples/deck/yarn.lock
+++ b/examples/deck/yarn.lock
@@ -2,11 +2,11 @@
 # yarn lockfile v1
 
 
-"@deck.gl/core@^6.0.0-beta.1":
-  version "6.0.0-beta.1"
-  resolved "https://registry.yarnpkg.com/@deck.gl/core/-/core-6.0.0-beta.1.tgz#0a4cd92da74acf43367f98c7953c089fca09a220"
+"@deck.gl/core@^6.0.1":
+  version "6.0.1"
+  resolved "https://registry.yarnpkg.com/@deck.gl/core/-/core-6.0.1.tgz#a563dfbbc631f2d7436b97fd53f98f3836c1d46a"
   dependencies:
-    luma.gl "^6.0.0-beta.1"
+    luma.gl "^6.0.0"
     math.gl "^2.0.0"
     mjolnir.js "^1.0.0"
     probe.gl "^1.0.0"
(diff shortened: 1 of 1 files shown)


georgios-uber:
@matthrice You need to rebase the branch. This is not based on master.

examples/deck/example.js
```diff
@@ -237,6 +259,28 @@ export default class Example extends Component<
...
+        },
+        existing: {
+          x: 58,
+          y: 0,
+          width: 58,
+          height: 58,
+          mask: false
+        }
+      },
+      getIcon: d => (d.type === 'existing' ? 'existing' : 'intermediate'),
```
supersonicclay: could this just be `d => d.type`?

docs/api-reference/layers/editable-geojson-layer.md
```diff
@@ -216,6 +216,43 @@ Edit handles are the points rendered on a feature to indicate interactive capabi
 
 * Default: `handle => (handle.type === 'existing' ? 5 : 3)`
 
+### Edit Handles (Icons)
```
supersonicclay: How about the following?

* `editHandleType`: `'point'|'icon'`
* `editHandlePoint...`: are just proxies for the `ScatterplotLayer` props
* `editHandleIcon...`: are just proxies for the [`IconLayer` props](http://deck.gl/#/documentation/deckgl-api-reference/layers/icon-layer?section=properties). So:
  * `editHandleIconAtlas'
  * `editHandleIconMapping`
  * `editHandleIconSizeScale`
  * These accessors would receive an edit handle object as an argument)
  * `editHandleIconGetIcon`
  * `editHandleIconGetSize`
  * `editHandleIconGetColor`
  * `editHandleIconGetAngle`

And I think it'd be cleaner to just have two separate props for getting color, size, and scale rather than have the same prop control two different things, especially if it is mis-named like `getEditHandlePointRadiusScale`.

matthrice: Agreed, this all makes sense.

PR #40, opened 2018-07-25
Downgradable geometry

supersonicclay:
With this functionality, a user can continue to remove points until they're just left with a single point (either a `Point` or `MultiPoint` feature). I call this "downgradable" geometries.

Of course, a consuming application can always choose to reject the edit if, for instance, a position is attempting to be removed from a triangle `Polygon` (as that would downgrade it to a `LineString`).


supersonicclay:
I think we could add that in the future. For now, users can decline an upgrade/downgrade in their `onEdit` callback by inspecting the geometry type before and after the edit.

Issue #53, opened 2018-08-27
Add creatCircle2 mode


georgios-uber:
Create a circle by defining diameter and automatically put the center.

PR #55, opened 2018-08-28
Add Create Circle by Bounding Box

bansalra:


georgios-uber:
#53

supersonicclay:
I wonder if this is just a personal preference of how to draw circles or if there is a compelling use case to justify adding extra modes.

My preference would be (in order):
1. don't introduce the complexity if it is just preference
2. choose a single way to draw circles and do that
3. name it something better than "2"

georgios-uber:
I'll take "3", what do you suggest?

supersonicclay:
After thinking about it more, I think this should be handled like other vector drawing tools do.

Rename `drawCircle` to `drawEllipse`. The user can customize the ellipse drawn with keyboard modifier keys.

* Draw ellipse using a bounding box will be the default (click at one corner, click at the other corner)
* Hold shift to constrain the ellipse to a circle
* Hold cmd (or ctrl on windows) to have the ellipse originate from the center point
* Hold shift+cmd (or shift+ctrl on windows) to draw a circle originating from the center point (as the `drawCircle` mode does today)

These are the same keyboard modifiers that gimp uses for its tools.

supersonicclay:
Hmm, maybe we should then have them be 3 separate modes then. How about these names?

* `drawEllipseByBoundingBox`
* `drawEllipseFromCenter`
* `drawCircleByBoundingBox`
* `drawCircleFromCenter`

It'd also be nice to provide some mechanism to default or "suggest" mode switching based on modifier keys. Perhaps some kind of "controller" class. But that could be added in the future.

georgios-uber:
Sounds great, ❤️ the names 🙈

examples/deck/example.js
```diff
@@ -166,7 +166,8 @@ export default class Example extends Component<
               <option value="drawLineString">drawLineString</option>
               <option value="drawPolygon">drawPolygon</option>
               <option value="drawRectangle">drawRectangle</option>
-              <option value="drawCircle">drawCircle</option>
+              <option value="drawCircleFromCenter">drawCircleFromCenter</option>
+              <option value="drawCircleByBoundingBox">drawCircleByBoundingBox</option>
```
georgios-uber: Add all 4 options.

supersonicclay:
Looks like documentation was missed on this one.

PR #62, opened 2018-08-30
add support for creating an individual point feature

jonpeters:
Basically, I just added a no-op mode of "drawPoint". It works because the first click when adding a feature creates a "Point" feature anyway. I could have done the same thing using "drawLineString", but it seems nebula will transiently render the dotted line for the "next" point in the line string, which is visible for a 1/2 second or so.

### DIFF
--- a/examples/deck/example.js
+++ b/examples/deck/example.js
@@ -92,14 +92,8 @@ export default class Example extends Component<
 
   _onLayerClick = info => {
     console.log('onLayerClick', info); // eslint-disable-line
-    const { mode } = this.state;
 
-    if (mode === 'drawPoint') {
-      // TODO why is info.index *always* 0 for a new feature?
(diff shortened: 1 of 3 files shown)


jonpeters:
@supersonicclay updated docs and examples

examples/deck/example.js
```diff
@@ -92,8 +92,14 @@ export default class Example extends Component<
 
   _onLayerClick = info => {
     console.log('onLayerClick', info); // eslint-disable-line
+    const { mode } = this.state;
 
-    if (this.state.mode !== 'view') {
+    if (mode === 'drawPoint') {
```
supersonicclay: For now, I don't think we need this check. A consuming application can do this kind of "bulk draw points" mode by leaving `selectedFeatureIndexes=[]` after the `onEdit` callback.

PR #66, opened 2018-09-04
Support for adding a new point to an existing feature anywhere on the…

jonpeters:
… line/polygon

Note, this PR is not yet complete; remaining items:

- add support for MultiLineString / MultiPolygon
- add a "hint" point
- add functionality to examples
- update documentation


package.json
```diff
@@ -51,13 +51,14 @@
...
+    "@turf/center": ">=4.0.0",
     "@turf/circle": ">=4.0.0",
     "@turf/destination": ">=4.0.0",
     "@turf/difference": ">=4.0.0",
     "@turf/distance": ">=4.0.0",
-    "@turf/helpers": ">=4.0.0",
-    "@turf/center": ">=4.0.0",
     "@turf/ellipse": ">=4.0.0",
+    "@turf/helpers": ">=4.0.0",
+    "@turf/nearest-point-on-line": "^6.0.2",
```
georgios-uber: Let's do `>=4.0.0` to keep everything consistent.

georgios-uber:
Please update `docs/api-reference/layers/editable-geojson-layer.md`

supersonicclay:
1. i think you should only render the hint if within X pixels of a line, but not if within Y of an existing point, and X and Y should be the same pixels that will take place when the user clicks or drags
2. i think users should be able to drag the point rather than needing to first click to add an intermediate point
3. i think we should render that hint point in a different color (i.e. it should have `type=='intermediate'` not `type='existing'`)

supersonicclay:
I think we can make this better tested by adding the functionality of getting the hint point to `EditableFeatureCollection.getEditHandles()`. You'd need to call it repeatedly on pointer move and it'd need to take an additional argument of the user's pointer position. But that'd get it out of the already-too-large `EditableGeoJsonLayer`.

src/lib/layers/editable-geojson-layer.js
```diff
@@ -575,6 +580,20 @@ export default class EditableGeoJsonLayer extends EditableLayer {
     }
   }
 
+  isPointerWithinHintPointMaxDistance(screenCoords: Array<any>, snapPoint: Object) {
+    const { hintPointMaxDistance } = this.props;
+    let isWithinMaxDistance = true;
+    if (hintPointMaxDistance) {
+      const snapPointScreenCoords = this.context.viewport.project(snapPoint.geometry.coordinates);
+      const distanceFromPointerToSnapPointPixels = math.distance(
```
supersonicclay: I don't think we need to bring in a whole new dependency just to do distance. Should be able to simply do:
```
side1 = abs(screenCoords[0]-snapPointScreenCoords[0])
side2 = abs(screenCoords[1]-snapPointScreenCoords[1])
distance = sqrt(side1^2+side2^2)
```

src/lib/layers/editable-geojson-layer.js
```diff
@@ -28,6 +29,8 @@ const defaultProps = {
   // Edit and interaction events
   onEdit: () => {},
 
+  hintPointMaxDistance: 0,
+  showHintPoint: true,
```
supersonicclay: I don't think this needs to be configurable. Just show it if and only if `mode=modify`.

src/lib/layers/editable-geojson-layer.js
```diff
@@ -28,6 +29,8 @@ const defaultProps = {
   // Edit and interaction events
   onEdit: () => {},
 
+  hintPointMaxDistance: 0,
```
supersonicclay: All new props need documentation. But I'm OK with this being a constant for now.

src/lib/layers/editable-geojson-layer.js
```diff
@@ -171,11 +176,14 @@ export default class EditableGeoJsonLayer extends EditableLayer {
       drawFeature = this.getDrawFeature(selectedFeature, props.mode, null);
     }
 
+    const hintPoint = selectedFeatures.length ? this.state.hintPoint : [];
```
supersonicclay: I think we should stick with the term "intermediate edit handle" rather than "hint point".

src/lib/layers/editable-geojson-layer.js
```diff
@@ -123,6 +126,7 @@ export default class EditableGeoJsonLayer extends EditableLayer {
 
     layers = layers.concat(this.createPointLayers());
     layers = layers.concat(this.createDrawLayers());
+    layers = layers.concat(this.createHintPointLayer());
```
supersonicclay: No need for a new layer, we already have an edit handle layer with callbacks that allow applications to customize edit handles (e.g. change the color, change to icons, etc).

src/lib/layers/editable-geojson-layer.js
```diff
@@ -383,21 +464,34 @@ export default class EditableGeoJsonLayer extends EditableLayer {
...
-    const { selectedFeatures } = this.state;
-    const editHandleInfo = this.getPickedEditHandle(picks);
-
-    if (!selectedFeatures.length) {
-      return;
-    }
+    // alternative method to forcing the user to click to add a new point;
+    // the following allows a drag initiation to create a new point and
+    // drag it without explicitliy selecting it
+    if (!this.getPickedEditHandle(picks) && this.state.hintPoint.length) {
```
supersonicclay: If you stick with the hint point being an "intermediate edit handle", you shouldn't need to change this code of adding intermediate positions.

src/lib/layers/editable-geojson-layer.js
jonpeters: Do you mind if we leave it for now? Our app's design spec doesn't call for a hint, despite it being a great feature.

PR #68, opened 2018-09-07
implement adding new point at any location on the line

jonpeters:


src/lib/editable-feature-collection.js
```diff
@@ -200,54 +209,124 @@ export class EditableFeatureCollection {
   /**
    * Returns a flat array of positions for the given feature along with their indexes into the feature's geometry's coordinates.
    *
-   * @param featureIndex The index of the feature to get edit handles
+   * @param featureIndexes The indexes of the features to get edit handles
+   * @param project Projection method to convert geographic coordinates to screen coordinates
+   * @param pointerPosition Object containg screen and ground coordinates for the location of the pointer
    */
-  getEditHandles(featureIndex: number) {
+  getEditHandles(featureIndexes: Array<number>, project?: Function, pointerPosition?: Object) {
```
supersonicclay: I think instead of getting a `project` function, screen coordinates, and ground coordinates, this should instead just get an array of picked features and a ground coordinate (no screen coordinates). The picked features would be the ones that are within deck.gl's picking threshold. Then `getEditHandles` only needs to look through those picks to see if any of them are lines or polygons in order to get the closest point along the lines. It can also short circuit the nearestPointOnLine check if one of those picks is an existing edit handle.

One benefit when someone has lots of selections. Consider someone has hundreds of features loaded and all are selected. In that case, you'd have to run `nearestPointOnLine` on thousands or millions of lines.

Another benefit is that the pixel picking threshold is in one place, deck.gl. This would simplify the picking logic and make it consistent. I noticed a slight discrepancy when dragging close enough to get the intermediate point to appear but far enough away that deck.gl doesn't consider the intermediate point clicked; in that case, the map pans rather than a point getting added.

Here's how you could achieve that. Move the call to `getEditHandles` from `EditableGeoJsonLayer.updateState` to `EditableGeoJsonLayer.onPointerMove`. And in `onPointerMove` do something like this:
```
const picks = this.context.layerManager.pickObject({
  x: screenCoords[0],
  y: screenCoords[1],
  mode: 'query',
  layers: [this.props.id],
  radius: 10,
  viewports: [this.context.viewport]
});
editableFeatureCollection.getEditHandles(props.selectedFeatureIndexes, picks, groundCoords);
```

src/lib/editable-feature-collection.js
```diff
@@ -200,54 +209,124 @@ export class EditableFeatureCollection {
...
-            );
-          }
+      }
+
+      // convert the linestring / coordinates to edit handles
+      handles = [...handles, ...getEditHandles(coordinates, positionIndexPrefix, featureIndex)];
+    });
+
+    // add the lone intermediate point
+    if (
```
supersonicclay: Looks like if your point is over an existing edit handle but also close enough to a line, the intermediate point is shown. We should only render an intermediate point if, and only if, dragging from that point will create a new intermediate point.

supersonicclay:
@jonpeters FYI, on master, `picks` is now pass to `onPointerMove`, so you should be able to use that.

jonpeters:
> @jonpeters FYI, on master, `picks` is now passed to `onPointerMove` from `EditableLayer`, so you should be able to use that.

@supersonicclay do these picks represent > 1 depth? The issue I was having before is that only 1 pick was ever returned, and sometimes it was the line and other times it was the edit handle.

supersonicclay:
> > @jonpeters FYI, on master, `picks` is now passed to `onPointerMove` from `EditableLayer`, so you should be able to use that.
> 
> @supersonicclay do these picks represent > 1 depth? The issue I was having before is that only 1 pick was ever returned, and sometimes it was the line and other times it was the edit handle.

You'll need to wait for that bug fix from deck.gl: https://github.com/uber/deck.gl/pull/2271

But once we upgrade nebula to include that fix from deck, you can change the pick call in `EditableLayer` rather than calling pick from `EditableGeoJsonLayer`.

PR #73, opened 2018-09-12
Feature/extend polyline from either end

jonpeters:
drawatfront


georgios-uber:
Interesting. @supersonicclay what do you think?

supersonicclay:
I like it.

Out of curiosity how would you see a consuming application toggle this prop? User holds down Ctrl/Cmd or based on which end the pointer is closest to?

I only ask because I wonder if that makes sense to be the default within nebula as well.

jonpeters:
> how would you see a consuming application toggle this prop?

@supersonicclay Our Map Maker application cares about which control point is "selected", so the prop is set based on the position index of the selected control point of the selected feature; if its the first edit handle in the array, the app sets the prop.

PR #74, opened 2018-09-13
Feature/enable remove position via keypress

jonpeters:
(Omitting a GIF since it would appear no differently than the default method of deleting an edit handle)

Basically, this is another usability requirement of the Map Maker application.

### DIFF
--- a/docs/api-reference/layers/editable-geojson-layer.md
+++ b/docs/api-reference/layers/editable-geojson-layer.md
@@ -179,6 +179,8 @@ The `onEdit` event is the core event provided by this layer and must be handled 
 
   * `movePosition`: A position was moved.
 
+  * `clickPosition`: A position was clicked.
+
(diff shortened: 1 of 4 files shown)


georgios-uber:
Seems good.

supersonicclay:
On the surface, this seems a bit too specific to require increasing the API's surface area. Why is this functionality needing to be built into the layer itself as opposed being implementing in the consuming application?

If there's some capability not exposed out of `EditableGeoJsonLayer` to enable this use case, perhaps we should look more closely at that before adding each application's concerns into nebula.

georgios-uber:
Agree with Clay.

jonpeters:
It doesn't have to be in the layer. I figured I'd put it there in case other consuming applications had the same requirement. Let me validate that it works when implemented on the application and then I will back out this PR ...

PR #77, opened 2018-09-19
Use Immutable.js

supersonicclay:
Simplify implementation of `EditableFeatureCollection` by using [immutable.js](https://facebook.github.io/immutable-js/).

Dependent on #79


georgios-uber:
We have decided against using immutable js in many places.
@ibgreen what do you think

georgios-uber:
Please split the flow in a different diff.

supersonicclay:
OK, I am seeing slower performance with immutable.js. I could look more at tuning it, but I was able to get things simplified some and was able to add the typings I hoped for in #79.

Abandoning this.

PR #78, opened 2018-09-19
Don't downgrade geometry type

supersonicclay:
The first part of the [upcoming tentative feature functionality](https://github.com/uber/nebula.gl/pull/71), we no longer want to have nebula change geometry type based on edits.

An example of how it works today:
* **Given**: Triangle polygon with a triangle hole inside it
* **When**: User clicks a point of the hole
* **Then**: Polygon no longer has a hole (because you can't have a Polygon with a LineString hole)
* **When**: User clicks a point of the outer ring
* **Then**: Polygon is downgraded to a LineString with two points (because you can't have a Polygon with only 2 points, and nebula assumes you want to downgrade)
* **When**: User clicks one of the points
* **Then**: LineString is downgraded to a Point
* **When**: User clicks the point
* **Then**: Nothing happens (nebula doesn't assume user wants to remove the feature entirely)

After this change, this is how it would work:
* **Given**: Triangle polygon with a triangle hole inside it
* **When**: User clicks a point of the hole
* **Then**: Polygon no longer has a hole (because you can't have a Polygon with a LineString hole)
* **When**: User clicks a point of the outer ring
* **Then**: Nothing happens (nebula doesn't assume user wants to downgrade the geometry type)

So, generally:
* a Point or MultiPoint will be simplified all the way down to a single point, but no further
* a LineString or MultiLineString will be simplified all the way down to a 2-point line, but no further
* a Polygon or MultiPolygon will be simplified all the way down to a triangle, but no further

Dependent on #79

### DIFF
--- a/docs/api-reference/layers/editable-geojson-layer.md
+++ b/docs/api-reference/layers/editable-geojson-layer.md
@@ -203,7 +203,7 @@ The `onEdit` event is the core event provided by this layer and must be handled 
 
   * `addPosition`: A position was added (either at the beginning, middle, or end of a feature's coordinates). Note: this may result in a feature being "upgraded" from one type to another (e.g. `Point` to `LineString` or `LineString` to `Polygon`).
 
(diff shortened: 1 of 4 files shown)


supersonicclay:
@georgios-uber: I'd like to land #79 first. Then this PR will simplify. If you want to review just these specific changes, they're in [this commit](https://github.com/uber/nebula.gl/pull/78/commits/5b7c58bf2b68278b851f2393a4ed8f0a9dcfd414)

supersonicclay:
@georgios-uber rebased

PR #79, opened 2018-09-19
Stronger typing

supersonicclay:


src/geojson-types.js
```diff
@@ -0,0 +1,75 @@
+// @flow
```
georgios-uber: Did you write this? I think there is a package you can import.

src/lib/layers/editable-geojson-layer.js
```diff
@@ -11,9 +11,9 @@ import destination from '@turf/destination';
 import bearing from '@turf/bearing';
 import pointToLineDistance from '@turf/point-to-line-distance';
 import { point, featureCollection as fc } from '@turf/helpers';
-import type { GeoJsonFeature } from '../../types';
-import { EditableFeatureCollection } from '../editable-feature-collection';
-import EditableLayer from './editable-layer';
+import type { Feature } from '../../geojson-types.js';
```
georgios-uber: Remove .js ! (Linter did not complain?)

supersonicclay: In my opinion (and several others), having `.js` is more standards-compliant.

Examples:

* Chrome requires the `.js` extension when using `<script type="module"></script>` [reference](https://stackoverflow.com/a/44481876/444917)
* Ryan Dahl regretted allowing omitting the extension for node.js requires [video](https://youtu.be/M3BM9TB-8yA?t=13m57s)

So, unless there's a compelling reason to leave them off, I'd vote to keep extensions.

src/geojson-types.js
supersonicclay: I looked at that, but nebula only supports a subset of GeoJSON (e.g. no `GeometryCollection`). Using our own types means we don't have to do `if (type !== 'GeometryCollection')` all over the place.

src/lib/layers/editable-geojson-layer.js
georgios-uber: OK.

src/geojson-types.js
georgios-uber: ok, great.

PR #80, opened 2018-09-20
 Convert EditableFeatureCollection to be mutable

supersonicclay:
But FeatureCollection it is editing is still immutable

### DIFF
--- a/src/test/lib/editable-feature-collection.test.js
+++ b/src/test/lib/editable-feature-collection.test.js
@@ -87,24 +87,24 @@ beforeEach(() => {
   };
 });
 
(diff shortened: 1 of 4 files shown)


src/test/lib/editable-feature-collection.test.js
```diff
@@ -480,21 +476,22 @@ describe('addPosition()', () => {
 
 describe('addFeature()', () => {
   it(`doesn't mutate original`, () => {
-    const features = new EditableFeatureCollection({
+    const leaveMeAlone = {
```
georgios-uber: 🤣

PR #91, opened 2018-09-26
add points anywhere on line (based on top of tentative feature refactor)

jonpeters:
This is the functionality to add new points anywhere on a line, using deck's picking engine instead of projects, and based on top of the "tentative" feature refactor.

### DIFF
--- a/modules/core/src/lib/editable-feature-collection.js
+++ b/modules/core/src/lib/editable-feature-collection.js
@@ -13,12 +13,12 @@ import { point, lineString as toLineString } from '@turf/helpers';
 
 import type {
   FeatureCollection,
-  Feature,
-  Geometry,
-  Point,
-  LineString,
-  Polygon,
-  Position
+    Feature,
+    Geometry,
+    Point,
+    LineString,
+    Polygon,
+    Position
 } from '../geojson-types.js';
 
 import { recursivelyTraverseNestedArrays } from './utils';
@@ -170,10 +170,7 @@ export class EditableFeatureCollection {
         );
         // tack on the lone intermediate point to the set of handles
         if (intermediatePoint) {
-          const {
-            geometry: { coordinates: position },
-            properties: { index }
-          } = intermediatePoint;
+          const { geometry: { coordinates: position }, properties: { index } } = intermediatePoint;
           handles = [
             ...handles,
             {
--- a/modules/core/src/lib/editable-feature-collection.js
+++ b/modules/core/src/lib/editable-feature-collection.js
@@ -170,7 +170,10 @@ export class EditableFeatureCollection {
         );
         // tack on the lone intermediate point to the set of handles
         if (intermediatePoint) {
-          const { geometry: { coordinates: position }, properties: { index } } = intermediatePoint;
+          const {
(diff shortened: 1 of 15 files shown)


modules/core/src/lib/layers/editable-layer.js
```diff
@@ -202,7 +202,8 @@ export default class EditableLayer extends CompositeLayer {
       mode: 'query',
       layers: [this.props.id],
       radius: 10,
-      viewports: [this.context.viewport]
+      viewports: [this.context.viewport],
+      depth: 2
```
jonpeters: @supersonicclay I noticed the same behavior as before, where the picked features alternated between including the intermediate edit handle, and thus produced a pulsating effect. Needed to still add the depth (and upgrade deck.gl) to resolve.

supersonicclay: Gotcha, we'll hold off until we can upgrade deck.gl to include the fix:
https://github.com/uber/deck.gl/pull/2271

modules/core/src/lib/editable-feature-collection.js
```diff
@@ -121,11 +123,59 @@ export class EditableFeatureCollection {
...
-          // Remove the last existing and 2 intermediate handles
-          handles = handles.slice(0, -3);
+          // Remove the last existing handle
+          handles = handles.slice(0, -1);
+        }
+      }
+    }
+
+    // intermediate edit handle
+    if (picks && picks.length && groundCoords) {
```
supersonicclay: Can we get test coverage of this as part of this PR?

modules/core/src/lib/editable-feature-collection.js
```diff
@@ -121,11 +123,59 @@ export class EditableFeatureCollection {
...
+      const existingEditHandle = picks.find(
+        pick => pick.isEditingHandle && pick.object && pick.object.type === 'existing'
+      );
+      // don't show intermediate point when too close to an existing edit handle
+      const featureAsPick = !existingEditHandle && picks.find(pick => !pick.isEditingHandle);
+
+      // is the feature in the pick selected
+      if (featureAsPick && this._selectedFeatureIndexes.includes(featureAsPick.index)) {
```
supersonicclay: Does this also need to check geometry type has lines? Something like `if (featureAsPick.geometry.type !== 'MultiPoint' && featureAsPick.geometry.type !== 'Point)`?

modules/core/src/lib/layers/editable-geojson-layer.js
```diff
@@ -320,8 +320,8 @@ export default class EditableGeoJsonLayer extends EditableLayer {
     }
   }
 
-  updateEditHandles() {
-    const editHandles = this.state.editableFeatureCollection.getEditHandles();
+  updateEditHandles(picks?: Array<Object>, groundCoords?: Position) {
```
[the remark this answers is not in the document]
jonpeters: @supersonicclay What about inside of updateState(), where those values aren't accessible?

supersonicclay:
@jonpeters looks good. rebase atop master just to make sure it still works with the translate feature that was recently added.

PR #93, opened 2018-10-01
Duplicate a feature

sivako:
In modify mode with action "duplicate", drag the selected feature will duplicated and move.
nebula-duplicate-feature


sivako:
> Agree with @supersonicclay
> Where is the PR for the fixed "Ability to transform translate (move) any feature" @sivako ?

@georgios-uber,  Here is the updated PR https://github.com/uber/nebula.gl/pull/95

georgios-uber:
Rebase

PR #95, opened 2018-10-03
Ability to move feature in new cursor mode

sivako:
@georgios-uber @supersonicclay, I have fixed the issues with move feature. 
Adding the new "cursor" mode,  In this mode the feature selections will have bounding box to feature.
With this user can move, duplicate, resize, (may be rotate) etc. (currently this PR will have "move" feature). 

Scenario: Select mode as "cursor" -> Select any feature -> mouse click + hold on feature and drag to move to desired location.

nebula-move-feature-2

### DIFF
--- a/docs/api-reference/layers/editable-geojson-layer.md
+++ b/docs/api-reference/layers/editable-geojson-layer.md
@@ -67,6 +67,8 @@ The `mode` property dictates what type of edits the user can perform and how to 
 
 * `view`: no edits are possible, but selection is still possible.
 
+* `cursor`: When user selected any feature, The bounding box will be drawn for the feature. Currently modeConfig.action: `transformTranslate` is available.
+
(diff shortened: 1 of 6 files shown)


georgios-uber:
I made a gif of how it works, looks like the shape is trying to avoid the mouse.
aaaaa

sivako:
Thanks you @supersonicclay and @georgios-uber.  I could see the issue mentioned and have fixed the issue.  Now the feature follows the mouse on pointer down + drag the feature.  I have move the "cursor" mode to different PR https://github.com/uber/nebula.gl/pull/98.  Can you please check both of these?. Thanks Again.

supersonicclay:
Hmmm. Still no.
translate

sivako:
Based on discussion with @supersonicclay, Will need to update feature scenario and logic behind. Also will raise PR after #96.

PR #97, opened 2018-10-03
WIP: Expose EditableFeatureCollection and allow it to be overridden

supersonicclay:


supersonicclay:
Abandoning in favor of #99

PR #98, opened 2018-10-04
New cursor mode, the selection will be bounding box

sivako:
The mode 'cursor' is added. With this mode when feature is selected, the bounding box will drawn. This will be useful to add more edit feature capabilities like move, rotate, expand and resize etc.
nebula-cursor-mode


sivako:
Yes, Agree, as we discussed this mode is not required for now.  I am closing this PR. Thanks @supersonicclay.

PR #103, opened 2018-10-09
Option to configure number of points for circle

bansalra:


docs/api-reference/layers/editable-geojson-layer.md
```diff
@@ -70,6 +70,7 @@ The `mode` property dictates which `ModeHandler` from the `modeHandlers` prop wi
 * Default: `null`
 
 An arbitraty object used to further configure the current `ModeHandler`.
+For `drawCircleByBoundingBox` mode, options can be passed to configure the circle steps.
```
supersonicclay: Document this is in the recently added mode handlers overview doc rather than here.

modules/core/src/lib/mode-handlers/draw-circle-by-bounding-box-handler.js
```diff
@@ -17,10 +17,19 @@ export class DrawCircleByBoundingBoxHandler extends TwoClickPolygonHandler {
       return result;
     }
 
+    const modeConfig = this.getModeConfig();
+    let { options = {} } = modeConfig;
+    if (options.steps < 16) {
+      console.warn(`Minimum steps to draw a circle is 16 `); // eslint-disable-line no-console,no-undef
```
supersonicclay: Is 16 the minimum in turf? If not, then we should make the minimum in nebula match turf. You can do further validation in your app if you want to make 16 the minimum.

bansalra: 4 is the minimum step required in turfjs. If set to 4, it will draw a rhombus. To look more like a circle configured it to 16.

Sure will update the changes

supersonicclay:
Can we do this for `drawCircleFromCenter` for consistency?

docs/api-reference/mode-handlers/overview.md
```diff
@@ -72,6 +72,12 @@ User can draw a new circular `Polygon` feature by clicking the center then along
 
 User can draw a new circular `Polygon` feature by clicking the two corners of bounding box.
 
+### ModeConfig
+
+A `modeConfig` object can be provided:
+
+* If options: `{steps: x}` means the circle will be drawn using `x` number of points.
```
supersonicclay: It's not clear if the provided object is `modeConfig: { options: { steps: 24 } }` or `modeConfig: { steps: 24 }`. Perhaps look at how `booleanOperation` is documented.

supersonicclay: And if `modeConfig: { options: { steps: 24 } }` is currently implemented, I think "config" and "options" is redundant. `modeConfig: { steps: 24 }` is simpler.

examples/deck/example.js
```diff
@@ -282,12 +282,22 @@ export default class Example extends Component<
       width: window.innerWidth
     };
 
+    let { modeConfig } = this.state;
+    if (mode === 'drawCircleByBoundingBox') {
```
supersonicclay: This will cause extra re-renders since you're creating a new object each time. Instead do `this.setState({ modeConfig: ... })`. See how `booleanOperation` is implemented.

docs/api-reference/mode-handlers/overview.md
bansalra: Options was added incase if there was a need to pass multiple configuration such as steps, unit etc rather than having a flattened object. 

Changed it for simplicity now as `modeConfig: { steps: 24 }`

PR #105, opened 2018-10-10
Ability to scale a feature

sivako:
New scale mode to increase/decrease the size of feature. 

Scenario: Select any feature, click+hold and drag away to increase. (click+hold and drag inward to decrease). 

nebula-scale

### DIFF
--- a/examples/deck/example.js
+++ b/examples/deck/example.js
@@ -198,6 +198,7 @@ export default class Example extends Component<
               <option value="view">view</option>
               <option value="modify">modify</option>
               <option value="rotate">rotate</option>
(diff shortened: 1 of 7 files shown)


examples/deck/example.js
```diff
@@ -198,6 +198,7 @@ export default class Example extends Component<
               <option value="view">view</option>
               <option value="modify">modify</option>
               <option value="rotate">rotate</option>
+              <option value="scale">scale</option>
```
supersonicclay: nit: also ignore `scaling` when logging from `onEdit`

PR #112, opened 2018-10-15
Specify 6.0.5 as minimum deck version

supersonicclay:


yarn.lock
```diff
@@ -2565,7 +2565,7 @@ decamelize@^1.0.0, decamelize@^1.1.1, decamelize@^1.1.2:
   version "1.2.0"
   resolved "https://registry.yarnpkg.com/decamelize/-/decamelize-1.2.0.tgz#f6534d15148269b20352e7bee26f501f9a191290"
 
-deck.gl@>=6.1.1:
+deck.gl@>=6.0.5:
   version "6.1.1"
```
georgios-uber: When 6.0.5 becomes available this should change here.

georgios-uber: What do you think?

modules/core/package.json
```diff
@@ -45,7 +45,7 @@
     "test-rendering": "(cd test/rendering-test && webpack-dev-server --config webpack.config.test-rendering.js --progress --hot --open)"
   },
   "dependencies": {
-    "@deck.gl/experimental-layers": ">=6.1.1",
+    "@deck.gl/experimental-layers": ">=6.0.5",
```
georgios-uber: Seems like experimental-layers should be left at 6.0.3

PR #113, opened 2018-10-16
Upgrade nebula turf dependencies to the latest.

romaxa:


supersonicclay:
Why? Since it is `>=`, a consuming application can use turf 6 without this change.

romaxa:
> Why? Since it is `>=`, a consuming application can use turf 6 today without any change to nebula.

good question,

supersonicclay:
I believe `>=` will jump over major versions. It is `~` and `^` that don't jump over major versions.

This is an issue with how yarn installs dependencies and sub-dependencies. I get around this by using yarn resolutions.
See this: https://github.com/yarnpkg/yarn/issues/4986#issuecomment-410717295
And this: https://yarnpkg.com/lang/en/docs/selective-version-resolutions/

PR #116, opened 2018-10-22
Feature/barebones configurable composite layer

jonpeters:
Hi, this is a more barebones version of [Matt Rice's PR](https://github.com/uber/nebula.gl/pull/46/).

I removed all the accessor wrapping stuff, since speaking with @supersonicclay revealed that there is really no need for an explicit context (e.g. isSelected, etc.) as the accessors will be defined in scope where they have access to all the application state needed anyway.

I did make it flexible enough so that both individual props or entire layers could be overridden.

Longer term, I think the functionality is a better fit in deck.gl as per @ibgreen's RFC and I am happy to take ownership of that. In the short term, I am hoping to to merge this since it enables so many of the polyline styling requirements in [our internal project] (and we are currently dependent on a local tarball of nebula containing this very change).

### DIFF
--- a/examples/deck/example.js
+++ b/examples/deck/example.js
@@ -11,6 +11,8 @@ import { EditableGeoJsonLayer } from 'nebula.gl';
 import sampleGeoJson from '../data/sample-geojson.json';
 
 import iconSheet from '../data/edit-handles.png';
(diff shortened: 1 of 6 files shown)


georgios-uber:
Why did you close this one @jonpeters ?

PR #121, opened 2018-10-26
Remove hacky event loop management

romaxa:
This code was done to workaround sync issues with custom react-mapbox component.
After switching to react-map-gl this code is not needed anymore.


georgios-uber:
Please add a photo to your profile 👍

PR #123, opened 2018-10-28
Disable hacky loop sync behind feature flag property

romaxa:
To provide backward compatibility it is better to expose feature property

### DIFF
--- a/modules/react/src/lib/nebula-react.js
+++ b/modules/react/src/lib/nebula-react.js
@@ -12,8 +12,10 @@ type Props = {
   viewport: Object,
   eventFilter?: Function,
   logger?: Object,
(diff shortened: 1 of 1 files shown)


romaxa:
Same property can be used to feature flag other hacks related to nebula events handling

modules/react/src/lib/nebula-react.js
```diff
@@ -12,8 +12,10 @@ type Props = {
   viewport: Object,
   eventFilter?: Function,
   logger?: Object,
+  enableLoopSync?: boolean,
```
georgios-uber: Add “hack” to the end.

georgios-uber:
@romaxa should I merge this?

PR #135, opened 2018-11-21
Implement Split Polygon

bansalra:
Split a polygon via line.
The gap between the polygons is configurable.

a

### DIFF
--- a/docs/api-reference/mode-handlers/overview.md
+++ b/docs/api-reference/mode-handlers/overview.md
@@ -148,8 +148,12 @@ User can split a polygon by drawing a new `LineString` feature on top of the pol
 The following options can be provided in the `modeConfig` object:
 * `gap` (optional):  `x <number>`
   * If gap: `x` means the spacing between the polygon would be `x`.
-* `unit` (optional):  `x <string>`
+  * Gap value should be greater than 0.
+  * Default gap is `0.1`
+
+* `unit` (optional):  `centimeters|feet|inches|meters|kilometers|miles|yards`
   * If unit: `x` means the unit used for the spacing would be of `x` unit.
+  * Default unit is `centimeters`
 
 ## Boolean Operations
 
--- a/docs/api-reference/mode-handlers/overview.md
+++ b/docs/api-reference/mode-handlers/overview.md
@@ -133,6 +133,24 @@ User can draw a new ellipse shape `Polygon` feature by clicking two corners of b
 
 User can draw a new ellipse shape `Polygon` feature by clicking center and two corners of the ellipse.
 
+## [SplitPolygonHandler](https://github.com/uber/nebula.gl/blob/master/modules/core/src/lib/mode-handlers/split-polygon-handler.js)
+
+* Mode name: `split`
+
+User can split a polygon by drawing a new `LineString` feature on top of the polygon.
+
+* If the first and the last click is outside the polygon, it will split the polygon
+
+* If the clicked position is inside the polygon, it will not split the polygon
+
+### ModeConfig
+
+The following options can be provided in the `modeConfig` object:
+* `gap` (optional):  `x <number>`
+  * If gap: `x` means the spacing between the polygon would be `x`.
+* `unit` (optional):  `x <string>`
+  * If unit: `x` means the unit used for the spacing would be of `x` unit.
+
 ## Boolean Operations
 
 For all polygon drawing modes, the following options can be provided in the `modeConfig` object:
--- a/modules/core/src/lib/mode-handlers/split-polygon-handler.js
+++ b/modules/core/src/lib/mode-handlers/split-polygon-handler.js
@@ -1,0 +1,113 @@
+// @flow
+
+import booleanPointInPolygon from '@turf/boolean-point-in-polygon';
+import turfDifference from '@turf/difference';
+import turfBuffer from '@turf/buffer';
+import type { Feature, Polygon } from '../../geojson-types.js';
+import type { ClickEvent, PointerMoveEvent } from '../event-types.js';
+import type { EditAction } from './mode-handler.js';
+import { ModeHandler } from './mode-handler.js';
+
+export class SplitPolygonHandler extends ModeHandler {
+  handleClick(event: ClickEvent): ?EditAction {
+    super.handleClick(event);
+
+    const editAction: ?EditAction = null;
+    const tentativeFeature = this.getTentativeFeature();
+    const selectedGeometry = this.getSelectedGeometry();
+    const clickSequence = this.getClickSequence();
+
+    if (!selectedGeometry) {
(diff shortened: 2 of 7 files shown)


supersonicclay:
Nice work. This will be a useful feature. Here are some observations about the functionality:

* Draw a line outside a polygon with a hole
  * Result: It adds a new polygon that fills the hole
  * Expected: change nothing
* Draw a line outside MultiPolygon
  * Result: splits the MultiPolygon into separate Polygons
  * Expected: change nothing

I think it would be less surprising if splitting didn't add new features, but instead just added polygons to existing features. This means that splitting a `Polygon` would upgrade it to a `MultiPolygon`. But splitting a `MultiPolygon` keeps it a `MultiPolygon`.

Can it work with multiple selected polygons? I'm ok if we do this in a subsequent PR.

docs/api-reference/mode-handlers/overview.md
```diff
@@ -127,6 +127,24 @@ User can draw a new ellipse shape `Polygon` feature by clicking two corners of b
 
 User can draw a new ellipse shape `Polygon` feature by clicking center and two corners of the ellipse.
 
+## [SplitPolygonHandler](https://github.com/uber/nebula.gl/blob/master/modules/core/src/lib/mode-handlers/split-polygon-handler.js)
```
supersonicclay: also document the new edit type

modules/core/src/lib/mode-handlers/split-polygon-handler.js
```diff
@@ -0,0 +1,113 @@
...
+    const modeConfig = this.getModeConfig();
+    // Default gap in between the polygon
+    const { gap = 0.001, units = 'miles' } = modeConfig;
+
+    const buffer = turfBuffer(tentativeFeature, gap, { units });
+    const updatedGeometry = turfDifference(selectedGeometry, buffer);
+    this._setTentativeFeature(null);
+    if (!updatedGeometry) {
+      // eslint-disable-next-line no-console,no-undef
+      console.warn('Canceling edit. Split Polygon erased');
```
romaxa: is it expected outcome on standard operation? do we really need console.warn here?

modules/core/src/lib/mode-handlers/split-polygon-handler.js
```diff
@@ -0,0 +1,113 @@
...
+    return result;
+  }
+
+  splitPolygon() {
+    const selectedGeometry = this.getSelectedGeometry();
+    const tentativeFeature = this.getTentativeFeature();
+    const featureIndex = this.getSelectedFeatureIndexes()[0];
+    const modeConfig = this.getModeConfig();
+    // Default gap in between the polygon
+    const { gap = 0.001, units = 'miles' } = modeConfig;
```
romaxa: is it possible to have gap = 0? why it is not default value? I would expect 2 polygons share same points if gap = 0.

modules/core/src/lib/mode-handlers/split-polygon-handler.js
```diff
@@ -0,0 +1,113 @@
...
+    const buffer = turfBuffer(tentativeFeature, gap, { units });
+    const updatedGeometry = turfDifference(selectedGeometry, buffer);
+    this._setTentativeFeature(null);
+    if (!updatedGeometry) {
+      // eslint-disable-next-line no-console,no-undef
+      console.warn('Canceling edit. Split Polygon erased');
+      return null;
+    }
+
+    const pieces = updatedGeometry.geometry.coordinates.map(coordinates => {
```
romaxa: will it handle source Multipolygon properly?

modules/core/src/lib/mode-handlers/split-polygon-handler.js
georgios-uber: Why default unit is not km? So 0.001 is one meter?

modules/core/src/lib/mode-handlers/split-polygon-handler.js
georgios-uber: We should have a system similar to deck.gl for logging, making a task for that.

georgios-uber: https://github.com/uber/nebula.gl/issues/136

modules/core/src/lib/mode-handlers/split-polygon-handler.js
georgios-uber: @bansalra You also need to consider polygons with holes, in which case coordinates will be an array of arrays, the first being the main polygon and the rest are the holes to punch through.

bansalra:
a

modules/core/src/lib/mode-handlers/split-polygon-handler.js
bansalra: Updated the code.

modules/core/src/lib/mode-handlers/split-polygon-handler.js
bansalra: Default Unit is now set to 0.1 centimeter (1mm). This way the polygons can share the same point

modules/core/src/lib/mode-handlers/split-polygon-handler.js
```diff
@@ -0,0 +1,113 @@
...
+
+import booleanPointInPolygon from '@turf/boolean-point-in-polygon';
+import turfDifference from '@turf/difference';
+import turfBuffer from '@turf/buffer';
+import type { Feature, Polygon } from '../../geojson-types.js';
+import type { ClickEvent, PointerMoveEvent } from '../event-types.js';
+import type { EditAction } from './mode-handler.js';
+import { ModeHandler } from './mode-handler.js';
+
+export class SplitPolygonHandler extends ModeHandler {
```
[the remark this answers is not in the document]
bansalra: Updated the test cases.

docs/api-reference/mode-handlers/overview.md
bansalra: Updated

docs/api-reference/mode-handlers/overview.md
```diff
@@ -127,6 +127,24 @@ User can draw a new ellipse shape `Polygon` feature by clicking two corners of b
...
+
+* If the first and the last click is outside the polygon, it will split the polygon
+
+* If the clicked position is inside the polygon, it will not split the polygon
+
+### ModeConfig
+
+The following options can be provided in the `modeConfig` object:
+* `gap` (optional):  `x <number>`
+  * If gap: `x` means the spacing between the polygon would be `x`.
```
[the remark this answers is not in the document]
bansalra: Updated the doc.

georgios-uber:
Great work @bansalra

bansalra:
Thank you @georgios-uber

Issue #142, opened 2018-12-11
Split polygon with 90 degree lock


romaxa:
#### Feature implementation

Add tool for splitting polygon with 90 degree angled polyline.

PR #143, opened 2018-12-11
Split 90 deg tool, fixing an issue (#142)

romaxa:
out

### DIFF
--- a/docs/api-reference/mode-handlers/overview.md
+++ b/docs/api-reference/mode-handlers/overview.md
@@ -161,6 +161,10 @@ The following options can be provided in the `modeConfig` object:
   * If unit: `x` means the unit used for the spacing would be of `x` unit.
   * Default unit is `centimeters`
 
+* `lock90Degree` (Boolean, optional)
+  * Default: `false`
+  * If true, all angles will be guaranteed to be 90 degrees.
+
 ## Boolean Operations
 
 For all polygon drawing modes, the following options can be provided in the `modeConfig` object:
--- a/examples/deck/example.js
+++ b/examples/deck/example.js
@@ -183,6 +183,10 @@ export default class Example extends Component<
   }
 
   _renderToolBox() {
+    const currentMode =
+      this.state.modeConfig && this.state.modeConfig.lock90Degree
+        ? 'split90Degree'
+        : this.state.mode;
     return (
       <div style={styles.toolbox}>
         <dl style={styles.toolboxList}>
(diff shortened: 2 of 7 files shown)


supersonicclay:
Would it make more sense for this to be a `modeConfig` option on the existing `spilt` mode? Seems like there would be a good deal of overlap.

romaxa:
Originally I tried to do it as the mode-config option but ran into the issue with nebula example heavily dependent on mode name and how it interacts with menu options. So I decided to not refactor that part, but just extend classes to avoid code duplication.

I can think about adding feature to distinguish tool options as toolName+Mode config parameters if you think we should do it in this CR, otherwise, I'll create another issue for nebula.

docs/api-reference/mode-handlers/overview.md
```diff
@@ -161,6 +161,9 @@ The following options can be provided in the `modeConfig` object:
   * If unit: `x` means the unit used for the spacing would be of `x` unit.
   * Default unit is `centimeters`
 
+* `lock90Degree`
```
supersonicclay: I like the way deck.gl does [its reference documentation](http://deck.gl/#/documentation/deckgl-api-reference/layers/layer?section=basic-properties). Specify name, type, optional/required, and default. Something like:

```
* `lock90Degree` (Boolean, optional)
  * Default: `false`

If true, all angles will be guaranteed to be 90 degrees.
```

examples/deck/example.js
```diff
@@ -218,6 +229,7 @@ export default class Example extends Component<
               <option value="draw90DegreePolygon">draw90DegreePolygon</option>
               <option value="drawRectangle">drawRectangle</option>
               <option value="split">split</option>
+              <option value="split90Degree">split90Degree</option>
```
supersonicclay: No longer needed

romaxa: This option is used to enable "split" tool with 90 degree mode enabled, why are you suggesting to remove it?

docs/api-reference/mode-handlers/overview.md
georgios-uber: @romaxa please update the doc like Clay suggested.

georgios-uber:
```suggestion
* `lock90Degree` (Boolean, optional)
  * Default: `false`

If true, all angles will be guaranteed to be 90 degrees.
```

PR #147, opened 2018-12-30
Allow using custom options object to use for supercluster instantiation

dewtea:


georgios-uber:
@dewtea Thanks for your contribution, please sign the CLA and then you should be able to merge this.

PR #150, opened 2019-01-17
Version 0.8.0

supersonicclay:


georgios-uber:
Shouldn't there be a big change for 0.8? What is it?

supersonicclay:
A method was renamed. Likely not used, but could be.

PR #156, opened 2019-01-29
Fix issue with clicking edit handle while drawing polygon

### LINKED ISSUE #141: Minimum polygon drawing example not working
#### Describe the bug
I'm trying to create a minimal working example to experiment with draw polygon.  However the code seems to work sporadically.  See code sandbox link below.

Mouse actions resulting in errors:
1. Immediately double-click
2. Click once, move mouse, then double-click

Mouse actions that work sporadically:
1. Draw at least 2 nodes, then double-click to complete the polygon
2. Draw at least 3 nodes, then click/double-click first node 

https://codesandbox.io/s/3192o0kjj5?module=%2Fsrc%2FApp.jsx

it seems bad json data is being passed to Deck.gl
deck gl error

nebula gl edit-error

#### To Do List 
  - [x] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

supersonicclay:
Fixes #141

### DIFF
--- a/modules/core/src/lib/mode-handlers/draw-polygon-handler.js
+++ b/modules/core/src/lib/mode-handlers/draw-polygon-handler.js
@@ -33,6 +33,12 @@ export class DrawPolygonHandler extends ModeHandler {
     let editAction: ?EditAction = null;
     const clickedEditHandle = getPickedEditHandle(picks);
 
+    if (clickedEditHandle) {
+      // remove the last element from the click sequence
+      const clickSequence = this.getClickSequence();
+      clickSequence.splice(clickSequence.length - 1, 1);
+    }
+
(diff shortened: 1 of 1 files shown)


modules/core/src/lib/mode-handlers/draw-polygon-handler.js
```diff
@@ -33,6 +33,12 @@ export class DrawPolygonHandler extends ModeHandler {
     let editAction: ?EditAction = null;
     const clickedEditHandle = getPickedEditHandle(picks);
 
+    if (clickedEditHandle) {
+      // remove the last element from the click sequence
```
romaxa: would be nice to add more info about why we have to remove it.

supersonicclay: Good idea

PR #160, opened 2019-02-09
Add support for simultaneously editing multiple polygons in translate, rotate, scale, duplicate modes

austintang:
The core logic of my change to add support for editing multiple features at once is to convert a list of selected features into a FeatureCollection with the feature indexes preserved in the geojson properties. Then the appropriate transformations are applied to the FeatureCollection (depending on what mode you're currently in). After the transformations are applied, the FeatureCollection is split up into individual features again.

### DIFF
--- a/modules/core/src/geojson-types.js
+++ b/modules/core/src/geojson-types.js
@@ -65,7 +65,7 @@ export type BoundingBoxArray = [number, number, number, number];
 export type FeatureOf<T: Geometry> = {
   type: 'Feature',
   geometry: T,
-  properties?: { index?: number },
+  properties?: { _internalIndex?: number },
   id?: string | number,
   bbox?: BoundingBoxArray
 };
--- a/modules/core/src/geojson-types.js
+++ b/modules/core/src/geojson-types.js
@@ -20,32 +20,38 @@ export type AnyCoordinates =
 
 export type Point = {
   type: 'Point',
-  coordinates: PointCoordinates
+  coordinates: PointCoordinates,
+  properties?: { [key: string]: any }
 };
 
 export type LineString = {
   type: 'LineString',
-  coordinates: LineStringCoordinates
+  coordinates: LineStringCoordinates,
+  properties?: { [key: string]: any }
 };
 
 export type Polygon = {
   type: 'Polygon',
-  coordinates: PolygonCoordinates
+  coordinates: PolygonCoordinates,
+  properties?: { [key: string]: any }
 };
 
 export type MultiPoint = {
   type: 'MultiPoint',
-  coordinates: MultiPointCoordinates
+  coordinates: MultiPointCoordinates,
+  properties?: { [key: string]: any }
 };
 
 export type MultiLineString = {
   type: 'MultiLineString',
-  coordinates: MultiLineStringCoordinates
+  coordinates: MultiLineStringCoordinates,
+  properties?: { [key: string]: any }
 };
 
 export type MultiPolygon = {
   type: 'MultiPolygon',
-  coordinates: MultiPolygonCoordinates
+  coordinates: MultiPolygonCoordinates,
+  properties?: { [key: string]: any }
 };
 
 export type Geometry = Point | LineString | Polygon | MultiPoint | MultiLineString | MultiPolygon;
@@ -59,7 +65,7 @@ export type BoundingBoxArray = [number, number, number, number];
 export type FeatureOf<T: Geometry> = {
   type: 'Feature',
   geometry: T,
-  properties?: {},
+  properties?: { index?: number },
   id?: string | number,
   bbox?: BoundingBoxArray
 };
--- a/modules/core/src/geojson-types.js
+++ b/modules/core/src/geojson-types.js
@@ -20,38 +20,32 @@ export type AnyCoordinates =
 
 export type Point = {
   type: 'Point',
-  coordinates: PointCoordinates,
-  properties?: { [key: string]: any }
+  coordinates: PointCoordinates
 };
 
 export type LineString = {
   type: 'LineString',
-  coordinates: LineStringCoordinates,
-  properties?: { [key: string]: any }
+  coordinates: LineStringCoordinates
 };
 
 export type Polygon = {
   type: 'Polygon',
-  coordinates: PolygonCoordinates,
-  properties?: { [key: string]: any }
+  coordinates: PolygonCoordinates
 };
 
 export type MultiPoint = {
   type: 'MultiPoint',
-  coordinates: MultiPointCoordinates,
-  properties?: { [key: string]: any }
+  coordinates: MultiPointCoordinates
 };
 
 export type MultiLineString = {
   type: 'MultiLineString',
-  coordinates: MultiLineStringCoordinates,
-  properties?: { [key: string]: any }
+  coordinates: MultiLineStringCoordinates
 };
 
 export type MultiPolygon = {
   type: 'MultiPolygon',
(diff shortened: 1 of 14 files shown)


georgios-uber:
LGTM
@austintang Can you update the docs?
@supersonicclay Do you have any comments?

georgios-uber:
@lxe feel free to add your comments

austintang:
@georgios-uber Sure I'll update the docs.

supersonicclay:
@austintang i noticed you can now click and drag anywhere on the map. Originally we required that a geometry be clicked and dragged in order to activate rotate, translate, scale. That way, users can pan the map while in those modes.

austintang:
@supersonicclay Got it. I'll make the appropriate changes.

modules/core/src/lib/mode-handlers/mode-handler.js
```diff
@@ -29,7 +30,7 @@ export type EditHandle = {
 export type EditAction = {
   updatedData: FeatureCollection,
   editType: string,
-  featureIndex: number,
+  featureIndex: number | Array<number>,
```
supersonicclay: I think it is strange to have `featureIndex` sometimes be a single number and sometimes an array. To support this and other edits in the future, we should probably rename to `featureIndexes` and always be an array. This would mean changing the existing modes to pass an array.

We also may want to do some future-proofing with `positionIndexes` and `position` as well.

How about this structure?
```
export type EditAction = {
  updatedData: FeatureCollection,
  editType: string,
  featureIndex: number[],
  editContext: ?any // For some edits this will contain positionIndexes, sometimes position. But nebula won't dictate the particular properties. This is can be fully dynamic and populated by the particular mode, or some mode in the future.
```

If we go this route, we should probably do it in its own PR and document the breaking change in the changelog.

austintang: @supersonicclay I fully agree with you here. The reason I did this was to make as little impact to existing code as possible. I'll create a follow up pull request for the featureIndexes change.

supersonicclay: I actually think the tweak to `EditAction` should come *before* this PR. But I'm want to see if @lxe and/or @georgios-uber has concern with the breaking change first.

supersonicclay:
I also would like to see some tests for this new feature.

modules/core/src/lib/mode-handlers/mode-handler.js
austintang: When you mentioned moving positionIndexes and position into an editContext object, were you talking about just talking about the EditAction type, or did you also want changes made to the EditHandle type?

supersonicclay:
I opened a PR here
#161

supersonicclay:
@austintang #161 is merged. rebase away!

austintang:
I'm surprised that there aren't any existing test cases for translate, duplicate, rotate and scale mode-handlers. I'll write test cases to account for the original use cases, in addition to adding cases for the new multi-select functionality.

austintang:
@supersonicclay I've added 72 unit tests that cover both single-feature and multi-feature actions for translate, rotate, duplicate and scale modes.

modules/core/src/geojson-types.js
```diff
@@ -59,7 +65,7 @@ export type BoundingBoxArray = [number, number, number, number];
 export type FeatureOf<T: Geometry> = {
   type: 'Feature',
   geometry: T,
-  properties?: {},
+  properties?: { index?: number },
```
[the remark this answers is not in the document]
austintang: @supersonicclay I addressed this issue in my latest diff. I also added test cases to ensure that properties provided by nebula users are preserved. I also renamed index to _internalIndex to avoid key collisions.

modules/core/src/lib/utils.js
```diff
@@ -87,3 +87,35 @@ export function generatePointsParallelToLinePoints(
...
+}
+
+export function convertFeatureCollectionToFeatureList({ features }: FeatureCollection): Array<any> {
+  return features.map(({ geometry, properties }) => {
+    // Turf functions do not preserve a properties value of _internalIndex: 0
+    // It is important for the properties._internalIndex value to be present to prevent
+    // a strange mass duplication bug.
+    if (properties) {
+      geometry.properties = properties;
+      if (!properties._internalIndex) {
```
supersonicclay: I think we can get rid of the need for `internalIndex`. And the reason is because the ordering of indexes in `selectedFeatureIndexes` is the same as the ordering of features in the `FeatureCollection` we hand to turf. And all these turf operations (rotate, scale, translate) should not be altering the ordering of indexes.

I simplified the implementation for `TranslateHandler` to show what I'm talking about. See my branch [here](https://github.com/austintang/nebula.gl/compare/master...uber:clay/review-160?expand=1).

austintang: This is much cleaner! However is it safe to rely on the order of `selectedIndexes` and `movedFeatures` or is order already accounted for?
```
    const movedFeatures = turfTransformTranslate(
      this._geometryBeforeTranslate,
      distanceMoved,
      direction
    );

    const selectedIndexes = this.getSelectedFeatureIndexes();
    for (let i = 0; i < selectedIndexes.length; i++) {
      const selectedIndex = selectedIndexes[i];
      const movedFeature = movedFeatures.features[i];
      updatedData = updatedData.replaceGeometry(selectedIndex, movedFeature.geometry);
    }
```

austintang: @supersonicclay I used your simplified implementation in my latest diff. Let me know if you see any other issues.

austintang:
@supersonicclay This PR is quite large. I am going to split this into two PRs. One will contain test cases for existing translate, rotate, duplicate, scale functionality, and the second will contain covered multi-select functionality. The PR with unit tests for existing functionality should be merged first. I will remove existing functionality unit tests from this PR - this will only contain new multi-select features and their unit tests.

modules/core/src/lib/mode-handlers/mode-handler.js
```diff
@@ -70,6 +71,42 @@ export class ModeHandler {
     return null;
   }
 
+  getSelectedFeatures(): Array<Feature> {
+    const { features } = this.featureCollection.getObject();
+    return this._selectedFeatureIndexes.map(index => {
+      const feature = features[index];
+      const featureCopy: any = { ...feature };
+      // Preserve feature index in geojson properties
+      featureCopy.properties = { ...featureCopy.properties, _internalIndex: index };
```
supersonicclay: Do you still need `getSelectedFeatures()` and `_internalIndex`?

modules/core/src/geojson-types.js
```diff
@@ -59,7 +65,7 @@ export type BoundingBoxArray = [number, number, number, number];
 export type FeatureOf<T: Geometry> = {
   type: 'Feature',
   geometry: T,
-  properties?: {},
+  properties?: { _internalIndex?: number },
```
supersonicclay: Do you still need `_internalIndex`?

modules/core/src/lib/mode-handlers/mode-handler.js
austintang: No these are no longer needed. I removed `getSelectedFeatures()` and `getSelectedGeometries()` methods from `ModeHandler`.

modules/core/src/geojson-types.js
austintang: This is also no longer needed. I made sure that there are no more references to `_internalIndex`.

PR #161, opened 2019-02-13
More flexible EditAction type

supersonicclay:
Rename featureIndex to featureIndexes for EditAction
Allow arbitrary editContext to be supplied by mode handlers


georgios-uber:
Very good work. I like that you updated the docs too.

PR #170, opened 2019-02-22
Added unit tests for existing translate, duplicate, rotate, and scale mode handler functionality

austintang:
This PR only contains new unit tests and modifications to mode-handler test utilities.


supersonicclay:
@austintang I'm fine doing it as just one PR. I stamped #160. So, I think we can close this one.

PR #177, opened 2019-03-06
Add polygon snapping to translate mode

austintang:
To enable polygon snapping, set `enablePolygonSnapping = true` in `EditableGeoJsonLayer` props. This implementation of polygon snapping only works in translate mode and only works for polygon features (multipolygons are not supported). Additionally, multiple selection is not yet supported.

I have plans to expand this feature to support rotate mode and to automatically rotate the selected polygon when snapping to a different polygon that has an edge that doesn't match the angle of the selected polygon's edge.


modules/core/package.json
```diff
@@ -61,6 +61,8 @@
     "@turf/helpers": ">=4.0.0",
     "@turf/intersect": ">=4.0.0",
     "@turf/line-intersect": ">=4.0.0",
+    "@turf/meta": ">=4.0.0",
```
georgios-uber: Good 👍

austintang:
I've added polygon snapping controls when the translate mode is selected in the nebula example
Screen Shot 2019-03-13 at 3 41 54 PM

modules/core/src/lib/mode-handlers/mode-handler.js
```diff
@@ -1,5 +1,12 @@
 // @flow
 
+import turfBearing from '@turf/bearing';
```
supersonicclay: Looks like you're adding a lot of code here that is specific to translate snapping mode. I wonder if you could create a separate class to encapsulate this functionality. I think the only code that should be in `ModeHandler` is stuff that applies to all (or at least several) modes.

austintang: @supersonicclay The snapping feature now snaps at polygon vertexes instead of edge midpoints. If polygon snapping is enabled, intermediate edit handles will be rendered to indicate the point that is closest to the selected polygon. Additionally, I abstracted the snapping logic into a new SnapHandler class.

snap_visuals

supersonicclay:
I enhanced the "complex" sample data in the example that you can use to test the performance.
https://github.com/uber/nebula.gl/pull/190

examples/deck/example.js
```diff
@@ -334,6 +341,60 @@ export default class Example extends Component<
...
+
+        <ToolboxRow>
+          <ToolboxLabel>Snap strength</ToolboxLabel>
+          <ToolboxControl>
+            <input
+              type="range"
+              min="0.1"
+              max="1"
+              step="0.1"
+              value={snapStrengthValue}
```
supersonicclay: I think rather than using an arbitrary "strength" unit. What if this is simply snap "pixels" and is used as the argument to picking?

austintang: Sure this sounds like a good idea. Is there a way for me to project world latlong space to screen pixel space? I'm thinking of using a snap pixel threshold that remains constant throughout all zoom levels.

examples/deck/example.js
```diff
@@ -133,6 +133,13 @@ export default class Example extends Component<
     window.addEventListener('resize', this._resize);
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (this.state.mode === 'translate' && this.state.mode !== prevState.mode) {
+      /* eslint-disable-next-line react/no-did-update-set-state */
+      this.setState({ modeConfig: { enablePolygonSnapping: false, snapStrength: 0.2 } });
```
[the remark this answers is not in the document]
austintang: This is required for snapping to work since the enable polygon snapping checkbox is associated with translate mode. There is a specific edge case that will case unsnapping to break.

The edge case is this: snap two features together, select a different mode, then select translate mode again. Leave the enable polygon snapping unchecked and try to unsnap the two polygons. Note that this cannot be done.

The problem is that without snapStrength specified through modeConfig, the code cannot determine at what mouse move threshold unsnapping should occur.

One thing I can do is to make the snapping related controls universal, where they will persist through mode changes. What do you think?

examples/deck/example.js
supersonicclay: You don't need to project to world space. Just use the `snapPixels` value as the `radius` argument to `pickObject`. Or you might need to call `pickMultipleObjects`, don't recall the exact method name for picking multiple. See `modules/layers/src/layers/editable-layer.js`. If `pickObject` returns a snappable edit handle (the ones that are gray in the demo), then you use turf and ground coordinates to get the closest point on that polygon to the point being dragged.

examples/deck/example.js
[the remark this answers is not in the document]
supersonicclay: I'm not seeing the need for an "unsnap" function. I think the logic could be more stateless. It can just "snap" if `pickObject` returns a gray edit handle and will do normal translate if not.

Something like this:

```js
if (modeConfig.snappingEnabled) {
  const picks = pickObjects({radius: modeConfig.snapPixels});
  // snapSourceEditHandle would be the purple handle being dragged
  const snapSourceEditHandle = getEditHandleBeingDragged(picks);
  // snapToEditHandle would be the gray handle close to the cursor or null if nothing close to the cursor
  const snapToEditHandle = getEditHandleOfNonSelectedFeature(picks);
  if (snapToEditHandle) {
    // perform snapping
    const startPoint = getCoordinateFromEditHandle(snapSourceEditHandle);
    const endPoint = getCoordinateFromEditHandle(snapToEditHandle);
    editAction = this.getTranslateAction(
        startPoint,
        endPoint,
        'translating'
      )
  }
}

if (!editAction) {
  // translate normally
 editAction = this.getTranslateAction(
        event.pointerDownGroundCoords,
        event.groundCoords,
        'translating'
      )
}
```

modules/layers/src/mode-handlers/snap-handler.js
```diff
@@ -0,0 +1,304 @@
...
+import turfEnvelope from '@turf/envelope';
+import { point as turfPoint } from '@turf/helpers';
+import turfTransformTranslate from '@turf/transform-translate';
+
+import type { FeatureCollection, Position } from '../geojson-types.js';
+import { SortedList, SortType } from '../utils';
+import type { EditHandle } from './mode-handler';
+import { ModeHandler, getEditHandlesForGeometry } from './mode-handler';
+
+export class SnapHandler extends ModeHandler {
```
supersonicclay: I'm not sure we need this class still. If you look at the pseudocode I posted, I think it may be pretty trivial to get the gray coordinate close to the cursor and use the existing `getTranslateAction` to translate from where the current purple coordinate is to the where the gray coordinate is.

examples/deck/example.js
```diff
@@ -334,6 +341,60 @@ export default class Example extends Component<
...
+              type="checkbox"
+              checked={Boolean(
+                this.state.modeConfig && this.state.modeConfig.enablePolygonSnapping
+              )}
+              onChange={event => {
+                const initModeConfig = this.state.modeConfig;
+                const { snapStrength } = initModeConfig;
+                const modeConfig = {
+                  ...initModeConfig,
+                  enablePolygonSnapping: Boolean(event.target.checked)
```
supersonicclay: How about just `enableSnapping` instead of `enablePolygonSnapping`?

examples/deck/example.js
austintang: How do I get access to the deck class's `pickMultipleObjects` method from within `mode-handler`? I can only find `pickObject` and `pickObjects` through `this._context.layerManager`.

supersonicclay: I think picks are already handed to the mode handlers. But you might need to tweak the interface to support multi-depth picking since you'll need the gray edit handle underneath the purple one. We might be able to hard-code the picking depth to `2` for now. But in the future, we might need some way for a mode handler to ask the layer to do deeper picks if necessary for that particular mode.

modules/layers/src/mode-handlers/translate-handler.js
```diff
@@ -2,18 +2,179 @@
...
   _isTranslatable: boolean;
+  _updatedData: ImmutableFeatureCollection;
+  _pickedHandle: ?EditHandle;
+  _isSnapped: boolean;
+
+  _getEditHandlePicks(event: any): HandlePicks {
+    const { screenCoords } = event;
+    const { snapPixels } = this.getModeConfig() || {};
+
+    const picks = this._context.layerManager.pickObject({
```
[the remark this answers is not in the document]
austintang: I didn't want to increase the depth for all EditableLayers. Also, by picking again in translate-handler, I can specifically pick from the `geojson-point-edit-handles` layer.

modules/layers/src/mode-handlers/mode-handler.js
```diff
@@ -18,7 +18,7 @@ export type EditHandle = {
   position: Position,
   positionIndexes: number[],
   featureIndex: number,
-  type: 'existing' | 'intermediate'
+  type: 'existing' | 'intermediate' | 'snap'
```
georgios-uber: Can you extract this as a separate type?

modules/layers/src/mode-handlers/mode-handler.js
```diff
@@ -289,7 +289,11 @@ export function getIntermediatePosition(position1: Position, position2: Position
   return intermediatePosition;
 }
 
-export function getEditHandlesForGeometry(geometry: Geometry, featureIndex: number) {
+export function getEditHandlesForGeometry(
+  geometry: Geometry,
+  featureIndex: number,
+  editHandleType: 'existing' | 'intermediate' | 'snap' = 'existing'
```
georgios-uber: and use the new type here. otherwise very confusing to read.

modules/layers/src/mode-handlers/translate-handler.js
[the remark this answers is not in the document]
georgios-uber: Agree, we need custom radius and layer ids.

modules/layers/src/mode-handlers/translate-handler.js
```diff
@@ -2,18 +2,182 @@
 
 import turfBearing from '@turf/bearing';
```
georgios-uber: Move all of this to snap-translate-handler.js

modules/layers/src/utils.js
```diff
@@ -87,3 +87,73 @@ export function generatePointsParallelToLinePoints(
 
   return [p3.geometry.coordinates, p4.geometry.coordinates];
 }
+
+export const SortType = {
+  MIN: 'min',
+  MAX: 'max'
+};
+
+export class SortedList {
```
georgios-uber: Seems like this is not needed any more.

modules/layers/test/lib/mode-handlers/translate-handler.test.js
```diff
@@ -25,3 +41,319 @@ testHandleStartDragging(
 );
 
 testHandleStopDragging(modeName, TranslateHandler, featureCollection);
+
```
georgios-uber: also move this to new file

modules/layers/test/lib/utils.test.js
```diff
@@ -117,3 +134,36 @@ describe('generatePointsParallelToLinePoints()', () => {
     expect(p4).toEqual([-123.09803547871964, 36.254027457172775]);
   });
 });
+
+const distanceObjList = [100, 28, 3, 493, 75, 1, 921].map(num => ({ distance: num }));
```
georgios-uber: These can go too

modules/layers/src/mode-handlers/translate-handler.js
austintang: Do you want me to rename the class to SnapTranslateHandler also? And on top of that, do you want me to rename the `EditableGeoJsonLayer` default props `modeHandler` key from `translate` to `snapTranslate`?
```js
modeHandlers: {
  snapTranslate: new SnapTranslateHandler()
}
```
 I worry that the latter change will be a breaking change though.

PR #184, opened 2019-03-19
Composite Mode Handler for creating and modifying at the same time

georgios-uber:


georgios-uber:
composite

modules/core/src/lib/mode-handlers/composite-mode-handler.js
```diff
@@ -0,0 +1,84 @@
...
+  handlers: Array<ModeHandler>;
+  options: Object;
+
+  constructor(handlers: Array<ModeHandler>, options: Object = {}) {
+    super();
+    this.handlers = handlers;
+    this.options = options;
+  }
+
+  _coalesce<T>(callback: any => T, resultEval: ?(T) => boolean = null): T {
```
supersonicclay: can be `callback: ModeHandler => T` instead of `any`

modules/core/src/lib/mode-handlers/composite-mode-handler.js
```diff
@@ -0,0 +1,84 @@
...
+
+  handleStopDragging(event: StopDraggingEvent): ?EditAction {
+    return this._coalesce(handler => handler.handleStopDragging(event));
+  }
+
+  getTentativeFeature(): ?Feature {
+    return this._coalesce(handler => handler.getTentativeFeature());
+  }
+
+  getEditHandles(picks?: Array<Object>, groundCoords?: Position): EditHandle[] {
```
supersonicclay: how about unioning the edit handles?

georgios-uber: This will break the comparison `editHandles !== this.state.editHandles`
So I'm adding a todo for now.

PR #189, opened 2019-03-21
Create @nebula.gl/layers

georgios-uber:


modules/layers/src/index.js
```diff
@@ -0,0 +1,41 @@
+// @flow
+
+export { ArrowStyles, DEFAULT_ARROWS, MAX_ARROWS } from './style.js';
+
+// Layers
+export { default as EditableGeoJsonLayer } from './layers/editable-geojson-layer.js';
```
supersonicclay: So, will both of these work after this PR?

```js
const EditableGeoJsonLayer = require('nebula.gl')
const EditableGeoJsonLayer = require('@nebula.gl/layers')
```

georgios-uber: That is the idea 👍 
Similar to `deck.gl` and `@deck.gl/core`

georgios-uber: To be precise you will need to:
```
const { EditableGeoJsonLayer } = require('nebula.gl');
const { EditableGeoJsonLayer } = require('@nebula.gl/layers');
```
or
```
import { EditableGeoJsonLayer } from 'nebula.gl';
import { EditableGeoJsonLayer } from '@nebula.gl/layers';
```

PR #192, opened 2019-03-22
Fix HtmlOverlay to handle possible null children

austintang:


georgios-uber:
@austintang can you add a unit test for that case?

austintang:
@georgios-uber Sure, I'll take care of it.

PR #194, opened 2019-03-29
Update immutable-feature-collection.js to copy elevation values when replacing coordinates

georgios-uber:


modules/layers/src/immutable-feature-collection.js
```diff
@@ -191,6 +191,17 @@ export class ImmutableFeatureCollection {
   }
 }
 
+function getUpdatedPosition(updatedPosition: Position, previousPosition: Position): Position {
+  // This function checks if the updatedPosition is mission elevation
```
georgios-uber:
```suggestion
  // This function checks if the updatedPosition is missing elevation
```

modules/layers/src/immutable-feature-collection.js
```diff
@@ -191,6 +191,17 @@ export class ImmutableFeatureCollection {
   }
 }
 
+function getUpdatedPosition(updatedPosition: Position, previousPosition: Position): Position {
+  // This function checks if the updatedPosition is missing elevation
+  // and copies if from previousPosition
```
georgios-uber:
```suggestion
  // and copies it from previousPosition
```

PR #200, opened 2019-04-14
Implement elevation-aware nearestPointOnLine

georgios-uber:


georgios-uber:
Moved to https://github.com/uber/nebula.gl/pull/236

PR #217, opened 2019-05-06
use gatsby-remark-embedded-codesandbox for inline examples in docs

georgios-uber:


georgios-uber:
Screen Shot 2019-05-05 at 9 02 59 PM

georgios-uber:
@ibgreen 2 problems:
- How to resize?
- Where to store mapbox token?

ibgreen:
This is awesome. Happy to accept this as standard feature in ocular-gatsby! @jckr 

> How to resize?

This seems to be part of the markdown system, so any resize would be handled by those plugins. We'll need to read up more on how this works?

> Where to store mapbox token?

I believe @jianhuang01 implemented a token query in the build script for deck.gl. Let's check if that is easy to port.

PR #220, opened 2019-05-07
Ability to toggle GeoJSON text

supersonicclay:
Stringifying GeoJSON is super slow and cripples nebula's perf. Hide by default.

### DIFF
--- a/examples/deck/example.js
+++ b/examples/deck/example.js
@@ -122,7 +122,8 @@ export default class Example extends Component<
     pointsRemovable: boolean,
     selectedFeatureIndexes: number[],
     editHandleType: string,
-    selectionTool: ?string
+    selectionTool: ?string,
+    showGeoJson: boolean
   }
 > {
   constructor() {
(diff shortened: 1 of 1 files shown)


examples/deck/example.js
```diff
@@ -418,16 +420,35 @@ export default class Example extends Component<
...
-          <textarea
-            id="geo-json-text"
-            rows={5}
-            style={{ width: '100%' }}
-            value={JSON.stringify(this.state.testFeatures)}
-            onChange={event => this.setState({ testFeatures: JSON.parse(event.target.value) })}
-          />
-        </ToolboxControl>
+        {this.state.showGeoJson && (
+          <>
```
georgios-uber: > Note that many tools don’t support it yet so you might want to explicitly write <React.Fragment> until the tooling catches up.

georgios-uber: Considering this is an example users might open in their editor.

PR #227, opened 2019-05-13
Enable flow for react-map-gl-draw

xintongxia:


modules/react-map-gl-draw/package.json
```diff
@@ -34,6 +34,8 @@
   },
   "dependencies": {
     "@turf/helpers": "^6.1.4",
+    "@nebula.gl/edit-modes": "^0.13.0",
```
georgios-uber:
```suggestion
    "@nebula.gl/edit-modes": "0.13.0",
```

modules/react-map-gl-draw/src/types.js
```diff
@@ -0,0 +1,34 @@
+// @flow
+
+export type {
```
georgios-uber: why you need this? Can you just import the types from `@nebula.gl/edit-modes` directly?

PR #234, opened 2019-06-18
add billboard prop to editable-geojson-layer

georgios-uber:


amangijoe:
billboard feature when set to false doesn't work

georgios-uber:
`billboard` is not related to extrusion. It only affects the angle of extrusion.

amangijoe:
thanks @georgios-uber Is there a way that I can disable extrusion or do I have to programmatically ignore elevation value from geojson data?

supersonicclay:
I'd suggest creating a new issue and you can reference this PR in the description

PR #236, opened 2019-06-18
Implement elevation-aware nearestPointOnLine

georgios-uber:


georgios-uber:
> Add test in the following PR

Writing unit test right now.

PR #239, opened 2019-06-20
Update advanced example style to dark mode

georgios-uber:


supersonicclay:
Much better! Only tweak I could see is perhaps a little more prevalent indication of selection. With all the colors, it's a little hard to tell the slightly more vibrant one being selected.

georgios-uber:
@supersonicclay I plan to work on that next.

PR #242, opened 2019-06-27
Support pan gestures when not interacting with a map feature

yunyu:
This commit allows users to pan the map in the EDIT_VERTEX or drawing modes, but only when the click target is not a map feature, the user is not currently dragging a vertex, and there is currently no uncommitted point. I have tested this change with the example and it behaves as expected in all modes. 

drag-fix

### DIFF
--- a/modules/react-map-gl-draw/src/editor.js
+++ b/modules/react-map-gl-draw/src/editor.js
@@ -284,14 +284,14 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
     }
 
     this._events = {
-      anyclick: evt => this._onEvent(this._onClick, evt),
+      anyclick: evt => this._onEvent(this._onClick, evt, true),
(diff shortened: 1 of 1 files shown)


modules/react-map-gl-draw/src/editor.js
```diff
@@ -556,6 +559,18 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
     }
   };
 
+  _onPan = (evt: MjolnirEvent) => {
```
xintongxia: add a comment about the conditions of panning event not being forward to map.

xintongxia:
Thanks for your contribution. Looks good to me.

modules/react-map-gl-draw/src/editor.js
yunyu: Will do!

PR #246, opened 2019-06-29
Only permit drags on the currently selected feature

yunyu:
Previously, you would be able to drag a feature even if you initiated the drag on top of another feature, which feels slightly off (and does not match other map drawing tools).

Before:
drag-fix-2-before

After:
drag-fix-2-after

### DIFF
--- a/modules/react-map-gl-draw/src/editor.js
+++ b/modules/react-map-gl-draw/src/editor.js
@@ -378,7 +378,7 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
         if (vertexIndex >= 0) {
           // dragging vertex
           this._updateFeature(selectedFeature, 'vertex', { vertexIndex, lngLat });
-        } else {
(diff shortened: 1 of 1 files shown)


xintongxia:
@yunyu 

Thanks for helping improve the library. Published a patch `0.14.5` with this included.

xintongxia:
@yunyu when you contribute to `react-map-gl-draw` module, please add me as the reviewer, which will notify me and check in time.

yunyu:
@xintongxia will do! Have a few more PRs in mind...

ylin-nuro:
@xintongxia I just noticed that this fix is somewhat flawed, in that it breaks drags if the mouse lags behind the polygon area. I will come up with a better one that fixes the mousedown drag start sometime.

yunyu:
@xintongxia I noticed that this fix is somewhat flawed, in that it breaks feature drags if the mouse lags behind the feature area. I will come up with a better one that fixes the mousedown drag start sometime.

PR #247, opened 2019-07-01
support customized react-map-gl-draw-editor style

xintongxia:


modules/react-map-gl-draw/src/editor.js
```diff
@@ -28,6 +28,7 @@ type EditorProps = {
   selectedFeatureId: ?Id,
   mode: string,
   clickRadius: ?number,
+  containerStyle: ?any,
```
macrigiuseppe: can you use:
```
style: ?object
```

modules/react-map-gl-draw/src/editor.js
```diff
@@ -1132,14 +1133,16 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
 
   _renderEditor = () => {
     const viewport = (this._context && this._context.viewport) || {};
+    const { containerStyle } = this.props;
```
macrigiuseppe: can you change to:
```
const { style } = this.props;
```

modules/react-map-gl-draw/src/editor.js
```diff
@@ -1132,14 +1133,16 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
...
     const { width, height } = viewport;
 
     return (
       <div
         id="editor"
         style={{
           width,
-          height
+          height,
+          ...containerStyle
```
macrigiuseppe: and this will become:
```
...style
```

PR #251, opened 2019-07-04
Make dragging features more resilient to lagging feature element

yunyu:
#246 resulted in weird dragging behavior when the feature element lags behind the mouse (such as when the map is zoomed out). This fixes the behavior so that you can only initiate a drag on top of the currently selected feature, but `mousemove`s modify the feature regardless of the cursor position.

It is safe to keep the `_matchesFeature` check in `_onPan` as-is, as `panmove` events aren't fired if `stopImmediatePropagation` is called on `panstart`. I also verified that the behavior remains correct for lines/points.

Before:
drag-fix-before-redux

After:
drag-fix-after-redux

### DIFF
--- a/modules/react-map-gl-draw/src/editor.js
+++ b/modules/react-map-gl-draw/src/editor.js
@@ -344,7 +344,7 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
         startDragPos: { x, y },
         isDragging: true
       });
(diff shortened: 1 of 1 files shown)


yunyu:
@xintongxia It doesn't seem like I can add you as a reviewer, so I'm tagging you here

xintongxia:
Looks good to me.

PR #253, opened 2019-07-08
refactor react-map-gl-draw part 1 - break down mode handlers

xintongxia:
The intention of this PR is to break down react-map-gl-draw into different editing modes and make the render (file `editor.js`) and event capture (file `mode-handler.js`) to work well with the new `@nebula.gl/edit-mode` API.

Next steps
- Extend from the new `@nebula.gl/edit-mode`
- Add flow
- Add tests


xintongxia:
> @Xintong I think you should import the edit-mode module instead of copying all the mode handlers.

I have been trying to use `edit-mode` module, but some of `edit-mode` are not applied. I am open to converge to edit-module in a following up PRs.

modules/react-map-gl-draw/src/mode-handlers/mode-handler.js
```diff
@@ -0,0 +1,380 @@
...
+  draggingVertexIndex: -1,
+  startDragPos: null,
+  isDragging: false,
+  didDrag: false,
+
+  // intermediate mouse position when drawing
+  uncommittedLngLat: null
+};
+
+export default class ModeHandler {
```
supersonicclay: I think we should rename this to `EditMode` to be consistent with the new naming in the `edit-modes` module. We still have `ModeHandler`, but that will all be refactored to be "edit modes".

modules/react-map-gl-draw/src/mode-handler.js
```diff
@@ -0,0 +1,416 @@
...
+import { DRAWING_MODE, EDIT_TYPE, MODES } from './constants';
+import { getScreenCoords, isNumeric, parseEventElement } from './edit-modes/utils';
+import {
+  BaseMode,
+  EditingMode,
+  DrawPointMode,
+  DrawLineStringMode,
+  DrawRectangleMode,
+  DrawPolygonMode
+} from './edit-modes';
```
georgios-uber: Shouldn't this be `@nebula.gl/edit-modes` ?

xintongxia: This is step 1 of refactoring, I need to make sure the rendering of react-map-gl-draw can handle the APIs and data flow of `@nebula.gl/edit-mode`. After Clay finished his refactoring and landed his PRs, I'll make another pass to extend from there.

PR #254, opened 2019-07-11
React-map-gl-draw: propagating onSelect mouse event

macrigiuseppe:
This pull request allows developers to capture the mouse event every time a user selects a feature.

This is essential for Kepler.gl when we have to capture the selected feature and display a context menu near to the mouse position.

Signed-off-by: Giuseppe Macri <gmacri@uber.com>

### DIFF
--- a/modules/react-map-gl-draw/src/editor.js
+++ b/modules/react-map-gl-draw/src/editor.js
@@ -446,9 +446,7 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
       feature &&
       (this.props.mode === MODES.SELECT_FEATURE || this.props.mode === MODES.EDIT_VERTEX)
     ) {
-      this.props.onSelect({
-        selectedFeatureId: feature.id
-      });
+      this.props.onSelect({ selectedFeatureId: feature.id }, evt);
     }
   };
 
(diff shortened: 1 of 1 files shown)


modules/react-map-gl-draw/src/editor.js
```diff
@@ -446,9 +446,7 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
       feature &&
       (this.props.mode === MODES.SELECT_FEATURE || this.props.mode === MODES.EDIT_VERTEX)
     ) {
-      this.props.onSelect({
-        selectedFeatureId: feature.id
-      });
+      this.props.onSelect({ selectedFeatureId: feature.id }, evt);
```
xintongxia: If you mostly need the coordinates of the pointerdown position, How about

```
this.props.onSelect({
  selectedFeatureId: feature.id,
  mapCoords: [lat, lng],
  screenCoords: [x, y]
});
```

macrigiuseppe: I also need to understand whether the user right clicked on the current feature and by passing the mouse event i can do so. All clicks, including right ones, are going through the _onClick method, due to the anyclick bouding.

xintongxia: Then could use the following format, which will be consistent with mode handling in `react-map-gl-draw` after refactor.

```
this.props.onSelect({
  selectedFeatureId: feature.id,
  mapCoords: [lat, lng],
  screenCoords: [x, y],
  sourceEvent: evt
});
```

macrigiuseppe: Yeah that works. I will update the diff

PR #257, opened 2019-07-15
 Support deleting point of a feature

xintongxia:
delete-demo

### DIFF
--- a/modules/react-map-gl-draw/src/editor.js
+++ b/modules/react-map-gl-draw/src/editor.js
@@ -42,6 +42,8 @@ type EditorProps = {
 type EditorState = {
   features: ?Array<Feature>,
   selectedFeatureId: ?Id,
+  selectedVertexIndex: ?number,
+
   uncommittedLngLat: ?Position,
 
   hoveredFeatureId: ?Id,
@@ -76,6 +78,7 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
         : null,
 
       selectedFeatureId: -1,
+      selectedVertexIndex: -1,
 
       hoveredFeatureId: null,
       hoveredLngLat: null,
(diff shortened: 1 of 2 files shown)


xintongxia:
@yunyu

modules/react-map-gl-draw/src/editor.js
```diff
@@ -138,6 +144,36 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
   _events: any;
   _context: ?MapContext;
 
+  _onKeydown = (evt: MjolnirEvent) => {
+    const { features } = this.state;
+    if (
+      features &&
+      (evt.key === 'Backspace' || evt.key === 'Delete') &&
```
macrigiuseppe: i would recommend to user event.which https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/which in order to detect the key.
backspace: 8
delete: 46

I would create constants for those values

modules/react-map-gl-draw/src/editor.js
```diff
@@ -689,7 +720,7 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
     switch (index) {
       case hoveredVertexIndex:
         return RENDER_STATE.HOVERED;
-      case -1:
+      case 'uncommitted':
```
macrigiuseppe: create a constant for this value and you can also export to library consumers

modules/react-map-gl-draw/src/editor.js
```diff
@@ -65,6 +67,24 @@ const defaultProps = {
   onSelect: () => {}
 };
 
+const defaultEditingState = {
+  selectedFeatureId: -1,
```
macrigiuseppe: it looks like you are using __-1__ in several places, i would create a constant

modules/react-map-gl-draw/src/editor.js
```diff
@@ -295,7 +319,8 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
       panend: evt => this._onEvent(this._onPan, evt, false)
     };
 
-    this._context.eventManager.on(this._events, ref);
+    eventManager.on(this._events, ref);
+    eventManager.on('keydown', evt => this._onEvent(this._onKeydown, evt, false));
```
macrigiuseppe: why didn't you bind __keydown__ event as part of __this._events__ object?

supersonicclay:
In the past, we made a design choice that converting geometry types was unexpected (with the exception of going Polygon<=>MultiPolygon). So, in the `modify` edit mode of nebula, we will shrink a polygon down to a triangle, but no further. And when removing a point of a triangular hole, we remove the hole. And when removing a point of a triangle of a MultiPolygon, we'll remove the triangle.

See this for a demo of how it works in nebula.
delete-point

JSanchezIO:
Are there any updates on this?

PR #258, opened 2019-07-18
Simplify EditMode interface to be more stateless

supersonicclay:
## TODO

* [x] Refactor guides to be `FeatureCollection` instead of `ModeHandlerGuides`

### DIFF
--- a/examples/advanced/example.js
+++ b/examples/advanced/example.js
@@ -104,6 +104,11 @@ const POLYGON_DRAWING_MODES = [
   'drawEllipseUsing3Points'
 ];
 
+const EMPTY_FEATURE_COLLECTION = {
+  type: 'FeatureCollection',
+  features: []
+};
+
 const modeHandlers = Object.assign(
   {
     'drawLineString+modify': new CompositeModeHandler([
@@ -170,7 +175,7 @@ export default class Example extends Component<
 
     this.state = {
       viewport: initialViewport,
-      testFeatures: sampleGeoJson,
+      testFeatures: EMPTY_FEATURE_COLLECTION,
       mode: 'drawPolygon',
       modeConfig: null,
       pointsRemovable: true,
@@ -241,10 +246,8 @@ export default class Example extends Component<
       });
     } else if (type === 'blank') {
       this.setState({
-        testFeatures: {
-          type: 'FeatureCollection',
-          features: []
-        }
+        testFeatures: EMPTY_FEATURE_COLLECTION,
+        selectedFeatureIndexes: []
       });
     } else if (type === 'file') {
       const el = document.createElement('input');
@@ -734,9 +737,12 @@ export default class Example extends Component<
         if (
           !['movePosition', 'extruding', 'rotating', 'translating', 'scaling'].includes(editType)
         ) {
+          const updatedDataInfo = updatedData.features.map(
+            feature => `${feature.geometry.type}(${getPositionCount(feature.geometry)})`
+          );
           // Don't log edits that happen as the pointer moves since they're really chatty
           // eslint-disable-next-line
-          console.log('onEdit', editType, editContext);
+          console.log('onEdit', editType, editContext, JSON.stringify(updatedDataInfo));
         }
         if (editType === 'removePosition' && !this.state.pointsRemovable) {
           // This is a simple example of custom handling of edits
@@ -865,5 +871,28 @@ export default class Example extends Component<
         {this.state.featureMenu && this._renderFeatureMenu(this.state.featureMenu)}
       </div>
     );
+  }
+}
+
+function getPositionCount(geometry): number {
+  // const getSum = (total: number, num: number) => total + num;
+  const flatMap = (f, arr) => arr.reduce((x, y) => [...x, ...f(y)], []);
+
+  const { type, coordinates } = geometry;
+  switch (type) {
+    case 'Point':
+      return 1;
+    case 'LineString':
+    case 'MultiPoint':
+      return coordinates.length;
+    case 'Polygon':
+    case 'MultiLineString':
+      return flatMap(x => x, coordinates).length;
+    case 'MultiPolygon':
+      // return coordinates.reduce(getSum, 0);
+      return flatMap(x => flatMap(y => y, x), coordinates).length;
+    // return coordinates.reduce((acc, x) => acc.concat([x]), []).reduce(getSum, 0);
+    default:
+      throw Error(`Unknown geometry type: ${type}`);
   }
 }
--- a/examples/advanced/example.js
+++ b/examples/advanced/example.js
@@ -139,8 +139,16 @@ const FEATURE_COLORS = [
   'CCDFE5'
 ].map(hex2rgb);
 
-function getEditHandleColor(handle: Object) {
-  switch (handle.type) {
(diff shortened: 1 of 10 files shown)


modules/edit-modes/src/lib/edit-mode.js
```diff
@@ -1,118 +1,26 @@
...
-  onEdit: (editAction: EditAction<TData>) => void,
-
-  // Callback used to update guides
-  onUpdateGuides: (guides: ?TGuides) => void,
-
-  // Callback used to update cursor
-  onUpdateCursor: (cursor: string) => void
-};
-
 export interface EditMode<TData, TGuides> {
```
xintongxia: ```_guides: FeatureCollection```, // preserving last guides

supersonicclay: I think we want to keep `EditMode` more stateless and functional. So, I don't think we want to keep track of the last guides in the state of the mode. Also, `EditMode` is not specific to GeoJSON. For example, we may have an `EditMode` in the future for editing h3 cells.

Memoizing guides to avoid re-calculating them unnecessarily is something that will likely be specialized in each mode. For example, some modes will have guides based on the data alone, other modes need to take the pointer position into account.

examples/advanced/example.js
```diff
@@ -734,9 +737,12 @@ export default class Example extends Component<
         if (
           !['movePosition', 'extruding', 'rotating', 'translating', 'scaling'].includes(editType)
         ) {
+          const updatedDataInfo = updatedData.features.map(
+            feature => `${feature.geometry.type}(${getPositionCount(feature.geometry)})`
+          );
           // Don't log edits that happen as the pointer moves since they're really chatty
           // eslint-disable-next-line
-          console.log('onEdit', editType, editContext);
+          console.log('onEdit', editType, editContext, JSON.stringify(updatedDataInfo));
```
georgios-uber: Isn't this going to have a perf issue?

georgios-uber: Maybe add an on/off option to ui?

modules/layers/src/layers/editable-geojson-layer-edit-mode-poc.js
```diff
@@ -434,23 +404,28 @@ export default class EditableGeoJsonLayer_EDIT_MODE_POC extends EditableLayer {
...
+    this.getActiveModeHandler().handleStopDragging(event, this.getModeProps(this.props));
   }
 
   onPointerMove(event: PointerMoveEvent) {
-    this.getActiveModeHandler().handlePointerMove(event);
+    this.setState({ lastPointerMoveEvent: event });
+    this.getActiveModeHandler().handlePointerMove(event, this.getModeProps(this.props));
   }
 
   getCursor({ isDragging }: { isDragging: boolean }) {
```
xintongxia: When will `isDragging` state update?

modules/edit-modes/src/lib/geojson-edit-mode.js
```diff
@@ -300,67 +271,78 @@ export class BaseGeoJsonEditMode extends BaseEditMode<FeatureCollection, ModeHan
     return this.getAddFeatureAction(geometry);
   }
 
-  handleClick(event: ClickEvent): void {
-    const editAction = this.handleClickAdapter(event);
+  // TODO: factor out all the duplicate calls to setFeatureCollection
+
+  handleClick(event: ClickEvent, props: ModeProps<FeatureCollection>): void {
+    this.setFeatureCollection(props.data);
```
xintongxia: Why you need a local copy of `props.data` as `featureCollection`?

supersonicclay: `this.featureCollection` is an instance of `ImmutableFeatureCollection`. In a future diff, I'll make it more stateless. But there is lots of code in the `ModeHandler`s that does `this.getFeatureCollection()`. So, just trying to make the refactor a little easier.

modules/layers/src/layers/editable-geojson-layer-edit-mode-poc.js
supersonicclay: `EditableLayer` is the one that keeps track of that based on pointerdown and pointerup event.

examples/advanced/example.js
supersonicclay: We ignore the really chatty edits (e.g. ones that happen as the mouse is moving), so this should only happen when they click or when the start or stop dragging.

supersonicclay: We ignore the really chatty edits (e.g. ones that happen as the mouse is moving), so this should only happen when they click or when the start or stop dragging.

supersonicclay:
@xintongxia @georgios-uber this is ready for re-review.

@xintongxia The GeoJSON edit mode now uses FeatureCollection for the guides.

@georgios-uber: I had to make a breaking change to the elevate layer to keep the example app able to run both `ModeHandler`s and `EdiMode`s implementations.

examples/advanced/example.js
```diff
@@ -867,3 +905,26 @@ export default class Example extends Component<
...
+    case 'LineString':
+    case 'MultiPoint':
+      return coordinates.length;
+    case 'Polygon':
+    case 'MultiLineString':
+      return flatMap(x => x, coordinates).length;
+    case 'MultiPolygon':
+      // return coordinates.reduce(getSum, 0);
+      return flatMap(x => flatMap(y => y, x), coordinates).length;
+    // return coordinates.reduce((acc, x) => acc.concat([x]), []).reduce(getSum, 0);
```
xintongxia: ditto

examples/advanced/example.js
```diff
@@ -867,3 +905,26 @@ export default class Example extends Component<
...
+    case 'Point':
+      return 1;
+    case 'LineString':
+    case 'MultiPoint':
+      return coordinates.length;
+    case 'Polygon':
+    case 'MultiLineString':
+      return flatMap(x => x, coordinates).length;
+    case 'MultiPolygon':
+      // return coordinates.reduce(getSum, 0);
```
xintongxia: delete?

modules/edit-modes/src/types.js
```diff
@@ -63,3 +63,26 @@ export type PointerMoveEvent = {
   pointerDownMapCoords: ?Position,
   sourceEvent: any
 };
+
+export type ModeProps<TData> = {
```
xintongxia: Add another field `viewport`?

examples/advanced/example.js
supersonicclay: yep, removed in #222

examples/advanced/example.js
supersonicclay: yep, removed in #222

modules/edit-modes/src/types.js
supersonicclay: We can do that. How does this look?

```
export type Viewport = {
  longitude: number,
  latitude: number,
  zoom: number,
  bearing: number,
  pitch: number,
  width: number,
  height: number
};
```

Or do you need `project` and `unproject`?

xintongxia: I need `project` / `unproject`.

supersonicclay: ok. let's do that part in a separate PR

PR #260, opened 2019-07-19
Fix clickRadius and insert a point to line segment

xintongxia:
demo

### DIFF
--- a/modules/react-map-gl-draw/src/editor.js
+++ b/modules/react-map-gl-draw/src/editor.js
@@ -67,6 +67,8 @@ const defaultProps = {
   onSelect: () => {}
 };
 
+const UNCOMMITTED_ID = 'uncommitted';
+
 export default class Editor extends PureComponent<EditorProps, EditorState> {
   static defaultProps = defaultProps;
 
@@ -695,7 +697,7 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
     switch (index) {
       case hoveredVertexIndex:
         return RENDER_STATE.HOVERED;
-      case 'uncommitted':
+      case UNCOMMITTED_ID:
         return RENDER_STATE.UNCOMMITTED;
       default:
         return RENDER_STATE.INACTIVE;
@@ -712,7 +714,7 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
         return RENDER_STATE.SELECTED;
       case hoveredFeatureId:
         return RENDER_STATE.HOVERED;
-      case 'uncommitted':
+      case UNCOMMITTED_ID:
         return RENDER_STATE.UNCOMMITTED;
       default:
         return RENDER_STATE.INACTIVE;
--- a/modules/react-map-gl-draw/src/editor.js
+++ b/modules/react-map-gl-draw/src/editor.js
@@ -337,6 +337,7 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
     const { x, y } = this._getEventPosition(evt);
     const attributes = parseElemDataAttributes(evt.target);
 
+    // click segment
     if (attributes && attributes.type === ELEMENT_TYPE.VERTEX) {
       const { vertexIndex } = attributes;
       this.setState({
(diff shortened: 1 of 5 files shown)


georgios-uber:
👍 on the unit tests!

modules/react-map-gl-draw/src/editor.js
```diff
@@ -400,13 +398,19 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
           this._isLine(elem)
         ) {
           const [, , segmentId] = elem.id.split('.');
+          // segmentId is start vertexIndex
+          const vertexIndex = Number(segmentId);
```
macrigiuseppe: since you are using Number to make sure segmentId is an actual number, what would you do if the conversation fails (result is NaN)?

modules/react-map-gl-draw/src/editor.js
```diff
@@ -477,19 +493,25 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
...
+      let lngLat = uncommittedLngLat;
+      if (!lngLat) {
+        const { x, y } = this._getEventPosition(evt);
+        lngLat = this._unproject([x, y]);
+        lngLat = this._getClosestPositionOnSegment(vertexIndex, lngLat, feature);
+      }
 
       if (lngLat) {
-        const insertPosition = (Number(index) + 1) % feature.points.length;
+        const insertPosition = (Number(vertexIndex) + 1) % feature.points.length;
```
macrigiuseppe: you can just reuse vertexIndex variable

modules/react-map-gl-draw/src/editor.js
```diff
@@ -477,19 +493,25 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
       elem
     ) {
       const [, , segmentId] = elem.id.split('.');
-      const [index] = segmentId.split('-');
+      const vertexIndex = Number(segmentId);
```
macrigiuseppe: same as previous

modules/react-map-gl-draw/src/editor.js
macrigiuseppe: One suggestion I have to better handling segmentId, you can use the __data__ html property https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

modules/react-map-gl-draw/src/editor.js
```diff
@@ -1077,7 +1106,13 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
               <rect
                 key={`${index}.feature-hidden`}
                 id={`feature.${index}.hidden`}
-                style={{ ...style, fill: '#000', fillOpacity: 0 }}
+                style={{
+                  ...style,
+                  fill: '#000',
```
macrigiuseppe: define a constant

modules/react-map-gl-draw/src/editor.js
```diff
@@ -1077,7 +1106,13 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
...
+                  width: clickRadius,
+                  height: clickRadius,
+                  fillOpacity: 0
```
macrigiuseppe: same here

modules/react-map-gl-draw/src/editor.js
```diff
@@ -689,7 +695,7 @@ export default class Editor extends PureComponent<EditorProps, EditorState> {
     switch (index) {
       case hoveredVertexIndex:
         return RENDER_STATE.HOVERED;
-      case -1:
+      case 'uncommitted':
```
macrigiuseppe: I would create a constant

modules/react-map-gl-draw/src/style.js
```diff
@@ -22,7 +27,7 @@ export function getFeatureStyle({ feature, state }) {
         : state === RENDER_STATE.UNCOMMITTED || state === RENDER_STATE.CLOSING
           ? '#a9a9a9'
           : '#000',
-    strokeWidth: 4,
+    strokeWidth: 2,
```
macrigiuseppe: create a constant

PR #270, opened 2019-08-28
Improve react-map-gl-draw API

xintongxia:
Per discussion with @Pessimistress 

- Simply `react-map-gl-draw` API
- Add deleteFeature(s) to `ImmutableFeatureCollection`

API changes see `react-map-gl-draw.md`

With this PR, the minimum usage of `react-map-gl-draw` editor could be simplied as 
```
<Editor
  clickRadius={12}
  mode={selectedMode}
/>
```


docs/api-reference/react-map-gl-draw/react-map-gl-draw.md
```diff
@@ -10,14 +12,22 @@
...
-- `onUpdate` (Function, Required) - callback when anything is updated. Receives one argument `features` that is the updated list of GeoJSON features.
+- `features` (Feature[], Optional) - List of features in GeoJson format.
+- `selectedFeatureIndex` (String, Optional) - Index of the selected feature. `EditorModes` assigns a unique id to each feature which is stored in `feature.properties.id`.
+- `clickRadius` (Number, Optional) - Radius to detect features around a hovered or clicked point. Default value is `0`
 
-Feature object structure:
-`react-map-gl-draw` is stateful component.
+- `onSelect` (Function, Optional) - callback when a feature or an editHandle is selected. Receives an object containing the following parameters
+  - `selectedFeatureIndex`: selected feature index.
+  - `editHandleIndex`: selected editHandle index.
```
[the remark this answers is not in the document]
xintongxia: the main purpose is to support point (vertex) deletion and styling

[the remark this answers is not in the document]
supersonicclay: gotcha. In `ModifyMode`, vertexes are deleted by just clicking them. There's no selection required to delete. We should discuss how we want it to work. If selection is important, we can look at what it will take to implement in the `EditMode` interface.

examples/react-map-gl-draw/app.js
```diff
@@ -77,7 +77,9 @@ export default class App extends Component {
         <Editor
           ref={_ => (this._editorRef = _)}
           clickRadius={12}
-          onSelect={selected => this.setState({ selectedFeatureIndex: selected ? 0 : null })}
+          onSelect={selected => {
+            this.setState({ selectedFeatureIndex: selected && selected.selectedFeatureIndex });
```
Pessimistress: - When is `selected` ever null?
- Use `selectedFeature` instead?

xintongxia: - `selected` will be null when user click empty space while a feature is selected, or a polygon is finished. 
- why should use `selectedFeature`?

Pessimistress: If the callback argument also contains `mapCoords` and `screenCoords`, then it should always be a valid object even if nothing is selected.

modules/react-map-gl-draw/src/mode-handler.js
```diff
@@ -183,40 +212,72 @@ export default class ModeHandler extends Component<EditorProps, EditorState> {
...
+            selectedFeatureIndex: featureIndex,
+            screenCoords: pointerDownScreenCoords,
+            mapCoords: pointerDownMapCoords
           });
         } else {
-          onSelect({
-            selectedFeatureId: null,
-            selectedFeatureIndex: null
+          this._onSelect({
+            selectedFeature: null,
```
Pessimistress: Update documentation

docs/api-reference/react-map-gl-draw/react-map-gl-draw.md
```diff
@@ -66,72 +78,68 @@ Returns is a map of [style objects](https://reactjs.org/docs/dom-elements.html#s
...
-      selectedMode: EditorModes.READ_ONLY,
-      features: [],
-      selectedFeatureId: null
+      // map
+      viewport: DEFAULT_VIEWPORT,
+      // editor
+      selectedMode: EditorModes.READ_ONLY
     };
+    this._mapRef = null;
+    this._editorRef = null;
```
Pessimistress: What are the refs used for?

examples/react-map-gl-draw/app.js
xintongxia: I am not sure what you mean, but right now, `onSelect` will be triggered in the following cases
- a polygon is finished
- a feature is selected
- click empty space while a feature is selected

Pessimistress: According to your documentation, the callback parameter also provides information about the click itself, which should still be accessible even if no feature is selected.

docs/api-reference/react-map-gl-draw/react-map-gl-draw.md
```diff
@@ -1,23 +1,34 @@
...
   - `EditorModes.DRAW_POLYGON` - Lets you draw a GeoJson `Polygon` feature.
   - `EditorModes.DRAW_POINT` - Lets you draw a GeoJson `Point` feature.
   - `EditorModes.DRAW_RECTANGLE` - Lets you draw a `Rectangle` (represented as GeoJson `Polygon` feature).
 
-- `selectedFeatureId` (String, Optional) - id of the selected feature. `EditorModes` assigns a unique id to each feature which is stored in `feature.properties.id`.
-- `clickRadius` (Number, optional) - Radius to detect features around a hovered or clicked point. Default value is `0`
-
-- `onSelect` (Function, Required) - callback when a feature is selected. Receives an object containing `selectedFeatureId`.
-- `onUpdate` (Function, Required) - callback when anything is updated. Receives one argument `features` that is the updated list of GeoJSON features.
+- `features` (Feature[], Optional) - List of features in GeoJson format.
```
Pessimistress: Explain the difference between managed and stateful here

PR #283, opened 2019-09-20
Swap implementation to use edit-modes module

supersonicclay:


CHANGELOG.md
```diff
@@ -7,9 +7,16 @@ _All notable changes to this project will be documented in this file. This proje
 <!--
 ## Version TBD
 
-### Breaking changes between `EditableGeoJsonLayer` and `EditableGeoJsonLayerEditModePoc`
+### Breaking changes for `EditableGeoJsonLayer`
 
 * `featureIndexes` is now nested under `editContext.featureIndexes` in parameter passed to `onEdit` callback
+* Edit handle type is now under `properties.editHandleType` instead of `type` for edit handle styling accessors:
+  * getEditHandlePointColor
```
xintongxia: I am wondering how is nebula rendering edithandle now? is it using a deck.gl layer?

supersonicclay: Yep. It still uses deck.gl layer. It actually renders tentative feature and edit handles as a single deck.gl layer (`GeoJsonLayer`). The reason for this breaking change is that edit handles are now just GeoJSON features instead of their own custom object.

PR #284, opened 2019-09-23
Unwrap guide object when using _sublayerProps

supersonicclay:
Compatibility to deck.gl@7.2.

Workaround for https://github.com/uber/deck.gl/issues/3671


supersonicclay:
Just bumped the version used for the example. We'll keep compatibility with `deck.gl@^7.0.0`.

PR #296, opened 2019-10-22
Use deck.gl mesh layer

xintongxia:


modules/layers/test/lib/mode-handlers/draw-line-string-handler.test.js
```diff
@@ -81,8 +81,7 @@ describe('when no selection', () => {
       },
       featureIndexes: [featureCollection.features.length],
       editContext: {
-      featureIndexes: [featureCollection.features.length],
-
+        featureIndexes: [featureCollection.features.length]
```
georgios-uber: unrelated

xintongxia: yarn pretty did

xintongxia:
> Did you test the segment layer? Are the directional arrows showing?

@georgios-uber No, I don't have much context about the impact of this change. Is there something else I need verify?

georgios-uber:
> > Did you test the segment layer? Are the directional arrows showing?
> 
> @georgios-uber No, I don't have much context about the impact of this change. Is there something else I need verify?

try https://github.com/uber/nebula.gl/tree/master/examples/sf running this

georgios-uber:
Taking this over.

PR #299, opened 2019-11-11
Measure distance mode + tooltips

supersonicclay:
Adds ability for nebula to render tooltips. Adds a new `MeasureDistanceMode`. The new mode supports different units and the formatting of the tooltip text is customizable.

measure-distance

### DIFF
--- a/docs/api-reference/layers/editable-geojson-layer.md
+++ b/docs/api-reference/layers/editable-geojson-layer.md
@@ -248,11 +248,11 @@ Edit handle objects can be represented by either points or icons. `editHandlePoi
 
 #### `editHandlePointOutline` (Boolean, optional)
 
-* Default: `false`
+* Default: `true`
 
 #### `editHandlePointStrokeWidth` (Number, optional)
 
-* Default: `1`
+* Default: `2`
 
 #### `editHandlePointRadiusMinPixels` (Number, optional)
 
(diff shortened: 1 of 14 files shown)


Pessimistress:
Some comments on the UX/API:

- streetscape has been asked to provide a callback when the measurement changes
- `@turf/distance` only measures distance in 2D. The next release of deck.gl will support picking a 3D point.
- Another use case is to measure the angle by selecting 3 points -- possible to work it into this mode?
- It would be nice to be able to control the placement/styling of the tooltip

supersonicclay:
Callback when measurement changes would be good to add. Perhaps a callback can be handed in via `modeConfig`.

I think measure angle can be implemented as a separate mode that consuming applications can toggle as desired (e.g. toolbar button press, key modifier, etc).

Styling/placement of tooltip can be accomplished via `_subLayerProps` using `tooltips` as the id of the sublayer.

supersonicclay:
Ok, added the measurement callback functionality. I think the other measurements and 3d capabilities can be added in the future. What do you think @Pessimistress ?

haxzie:
@supersonicclay Any info on when this feature will be available? the latest relase on NPM is v0.16.0. Will it be available on v0.16.1?

leonardoviveiros:
@haxzie looks like they tried to release a new version with this feature, but the pipeline failed for other reason

https://github.com/uber/nebula.gl/runs/324055731

I'm too waiting for this to be realeased

supersonicclay:
This is available in 0.17.0+ And 0.17.2 has a measure area mode as well!

PR #300, opened 2019-11-14
Configuration option to disable dragging features

OrKoN:
We need to support a use case where dragging features in the edit mode is not allowed. I have added a parameter called `featuresDraggable` which is `true` by default. If `featuresDraggable === false`, dragging will be prevented in the editing mode.

Let me know if it's an interesting change for you and your thoughts on the API and missing pieces to make this change:

- is there any way to configure different modes? currently, `featuresDraggable` would only apply to the editing mode but other modes might have dragging as well
- what needs to be document and what tests would be needed make the change
- better way to disabled dragging?

### DIFF
--- a/modules/react-map-gl-draw/src/edit-modes/editing-mode.js
+++ b/modules/react-map-gl-draw/src/edit-modes/editing-mode.js
@@ -105,6 +105,9 @@ export default class EditingMode extends BaseMode {
       case ELEMENT_TYPE.FEATURE:
       case ELEMENT_TYPE.FILL:
       case ELEMENT_TYPE.SEGMENT:
+        if (!props.featuresDraggable) {
+          break;
+        }
         // dragging feature
         const dx = screenCoords[0] - lastPointerMoveEvent.screenCoords[0];
         const dy = screenCoords[1] - lastPointerMoveEvent.screenCoords[1];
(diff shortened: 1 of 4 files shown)


supersonicclay:
@xintongxia can you look at this one?

xintongxia:
Thanks for your contribution.

> is there any way to configure different modes? currently, featuresDraggable would only apply to the editing mode but other modes might have dragging as well

Right now react-map-gl-draw supports several drawing modes plus selecting and editing modes. In the drawing modes, dragging are not applied. 

So `featuresDraggable` is potentially useful for future editing modes.

> what needs to be document and what tests would be needed make the change
https://github.com/uber/nebula.gl/blob/master/docs/api-reference/react-map-gl-draw/react-map-gl-draw.md

About tests, now we don't have a good test coverage yet, we try to add tests when adding new features.

> better way to disabled dragging?

I am thinking to make react-map-gl-draw to accept customized editing mode, which you can have a customized editing mode that ignores dragging.

OrKoN:
Providing a completely custom mode would be good for customization in the future! Do you think adding the support for featuresDraggable makes sense at this point?

georgios-uber:
@OrKoN can you rebase-update this or if no longer needed just close.

georgios-uber:
@xintongxia can you review please?

georgios-uber:
@OrKoN can you push this again so the tests auto-run? I fixed the travisci issue.

OrKoN:
@georgios-uber it does not seem to work (requires FOSSA key). I have rebased again as well.

georgios-uber:
> @georgios-uber it does not seem to work (requires FOSSA key). I have rebased again as well.

OK... OK... now should be fixed, can you rebase and force-push?
(travis seems to not set the env vars when getting outside PRs).

PR #317, opened 2020-01-10
Fix deck 8

supersonicclay:
Fixes the integration with deck.gl 8. I took out the line dash props since they're handled differently in deck.gl 8. Also, I had to put the same tweak on the event handling as https://github.com/uber/deck.gl/blame/1a8b5d495b2e3f7b139ec3559002a010d73f329b/modules/core/src/lib/deck.js#L601

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,13 +4,26 @@ _All notable changes to this project will be documented in this file. This proje
 
 <!-- INSERT HERE -->
 
+<!--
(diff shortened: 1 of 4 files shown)


georgios-uber:
@supersonicclay can we forward 

> PathStyleExtension

 ?

supersonicclay:
I think we want to stick with pushing users to use `_subLayerProps` directly

PR #323, opened 2020-01-22
Implement ability to drag and draw in certain modes

### LINKED ISSUE #301: A flag to switch to standard draw mode
#### Feature request
Is it possible to introduce a flag for dragAndDraw, which is when true enable to draw rectange/polygon on mousedown and on releasing mouseclick (onmouseup) it would close the rectangle and fire onUpdate() method with the selected coords. I was looking at the code, and seems like it's possible. Just need a little guidance so that I can go ahead and create a pull request with that feature.

This feature will be helpful as it will introduce standard flow of drawing rectangle (mouseclick and drag which mouse click down and on release close drawing)

Thanks for this amazing package.

supersonicclay:
For all modes that implement TwoClickPolygonMode, support dragging to draw instead of needing two clicks. Two clicks is still the default.

dragToDraw

Fixes #301

TODO:
* [ ] Documentation
* [ ] Tests

### DIFF
--- a/docs/api-reference/modes/overview.md
+++ b/docs/api-reference/modes/overview.md
@@ -91,6 +91,13 @@ User can draw a new `Polygon` feature with 90 degree corners (right angle) by cl
 
 User can draw a new rectangular `Polygon` feature by clicking two opposing corners of the rectangle.
 
+### ModeConfig
+
+The following options can be provided in the `modeConfig` object:
+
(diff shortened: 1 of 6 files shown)


supersonicclay:
This is ready for review @georgios-uber.

supersonicclay:
@xintongxia you may want to review this also

Issue #289, opened 2019-10-03
Measurement tools


Pessimistress:
#### Feature request

Offer a set of tools (layers/modes) that allow the user to
- Draw line segment(s) and measure the distance
- Draw two line segments and measure the angle between them
- Draw a polygon and measure the area
- Draw an extruded polygon and measure the volume
- Draw a polygon and count rendered objects inside (e.g. point cloud)

Proof of concept: https://github.com/uber/streetscape.gl/pull/397

#### To Do List 
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Doc update
  - [ ] Test

supersonicclay:
Measure distance mode implemented in #299

leonardoviveiros:
Hello @supersonicclay,
I see that #299 introduced Measurement Mode on Demo website, but looks like the pipeline failed to generate a new version.

https://github.com/uber/nebula.gl/runs/324055731

I'd like to use nebula.gl with this MeasureDistanceMode. 
Is it possible to have a fix, and generate a new build version?

georgios-uber:
@leonardoviveiros the pipeline does not publish versions only verifies the build.
We will have a new release soon.

supersonicclay:
Measure area in #325

supersonicclay:
Measure angle in #332

PR #332, opened 2020-02-09
Ability to measure angles

supersonicclay:


xintongxia:
what is the use case for this mode?

modules/edit-modes/src/lib/measure-angle-mode.js
```diff
@@ -0,0 +1,130 @@
+// @flow
+
+import turfBearing from '@turf/bearing';
+import turfCenter from '@turf/center';
+import memoize from 'memoizee';
```
xintongxia: There is a small util file in deck which supports memorize, and avoid introduce another dependency.

https://github.com/uber/deck.gl/blob/master/modules/core/src/utils/memoize.js

modules/edit-modes/src/lib/measure-angle-mode.js
```diff
@@ -0,0 +1,130 @@
...
+
+export class MeasureAngleMode extends BaseGeoJsonEditMode {
+  _getTooltips = memoize((modeConfig, vertex, point1, point2) => {
+    let tooltips = DEFAULT_TOOLTIPS;
+
+    if (vertex && point1 && point2) {
+      const { formatTooltip, measurementCallback } = modeConfig || {};
+      const units = 'deg';
+
+      // const angle = turfAngle(point1, vertex, point2);
```
xintongxia: delete line 27?

supersonicclay:
> what is the use case for this mode?

We had a request for several measurement modes #289.

PR #340, opened 2020-02-24
Add transform edit mode

austintang:
This new `TransformMode` effectively merges translate, rotate, and scale modes into one, with the help of edit guides. This new mode also supports editing multiple selected features.
transform


modules/edit-modes/package.json
```diff
@@ -57,6 +57,7 @@
     "@turf/difference": ">=4.0.0",
     "@turf/distance": ">=4.0.0",
     "@turf/ellipse": ">=4.0.0",
+    "@turf/envelope": "^5.1.5",
```
georgios-uber: is this the same as bbox?

austintang: This is very similar to bbox, except turf envelope returns a bounding box feature instead of an array of 4 values. Do you want me to use bbox instead?

georgios-uber: Yes please if it is not a lot of work.

modules/edit-modes/src/lib/transform-mode.js
```diff
@@ -0,0 +1,319 @@
...
+  StopDraggingEvent,
+  DraggingEvent,
+  ModeProps,
+  EditHandleFeature
+} from '../types.js';
+import type { FeatureCollection, Position, Feature } from '../geojson-types.js';
+import { BaseGeoJsonEditMode, getIntermediatePosition } from './geojson-edit-mode';
+import { ImmutableFeatureCollection } from './immutable-feature-collection';
+
+export class TransformMode extends BaseGeoJsonEditMode {
```
supersonicclay: This mode needs tests as well. Look at draw-rectangle-mode or modify-mode tests as an example.

modules/edit-modes/src/lib/transform-mode.js
```diff
@@ -0,0 +1,319 @@
...
+    if (Array.isArray(features) && features.length === 1) {
+      const { type } = getGeom(features[0]);
+      if (type === 'Point') {
+        return true;
+      }
+    }
+    return false;
+  };
+
+  _getTranslateAction(
```
supersonicclay: Is there a way for this mode to use the translate, scale, and rotate modes inside itself rather than duplicating their logic?

modules/edit-modes/package.json
austintang: I updated my PR - I removed the `@turf/envelope` dependency and used bbox instead.

modules/edit-modes/src/lib/transform-mode.js
austintang: I updated my PR - scale and rotate modes now have edit handle guides. Translate mode functionality will remain unchanged. Transform mode's functionality will remain the same as what's depicted in GIF attached to this PR's description.

examples/advanced/example.js
```diff
@@ -15,6 +15,7 @@ import {
   type EditMode,
   ModifyMode,
   TranslateMode,
+  TransformMode,
```
supersonicclay: Can you also add this mode to the documentation at `docs/api-reference/modes/overview.md`?

modules/edit-modes/src/lib/scale-mode.js
```diff
@@ -1,101 +1,223 @@
...
-  DraggingEvent
+  DraggingEvent,
+  EditHandleFeature
 } from '../types.js';
-import { BaseGeoJsonEditMode, type GeoJsonEditAction } from './geojson-edit-mode.js';
+import { getPickedEditHandle } from '../utils';
+import { BaseGeoJsonEditMode } from './geojson-edit-mode.js';
 import { ImmutableFeatureCollection } from './immutable-feature-collection.js';
 
 export class ScaleMode extends BaseGeoJsonEditMode {
```
supersonicclay: Can you also add tests for scale mode? We had tests before at `modules/layers/test/lib/mode-handlers/scale-handler.test.js` that need ported over, but they probably need mostly rewritten since this mode works differently now.

modules/edit-modes/src/lib/rotate-mode.js
```diff
@@ -1,84 +1,195 @@
...
 import type { FeatureCollection, Position } from '../geojson-types.js';
-import { BaseGeoJsonEditMode, type GeoJsonEditAction } from './geojson-edit-mode.js';
+import {
+  BaseGeoJsonEditMode,
+  type GeoJsonEditAction,
+  getIntermediatePosition
+} from './geojson-edit-mode.js';
 import { ImmutableFeatureCollection } from './immutable-feature-collection.js';
 
 export class RotateMode extends BaseGeoJsonEditMode {
```
supersonicclay: Can you also add tests for rotate mode? We had tests before at `modules/layers/test/lib/mode-handlers/rotate-handler.test.js` that need ported over, but they probably need mostly rewritten since this mode works differently now.

modules/edit-modes/src/lib/rotate-mode.js
```diff
@@ -1,84 +1,195 @@
...
-  _isRotatable: boolean;
-  _geometryBeingRotated: ?FeatureCollection;
+  _selectedEditHandle: ?EditHandleFeature;
+  _selectedGeometry: ?FeatureCollection;
+  _previousMouseCoords: ?Position;
+  _isRotating: boolean = false;
+  _isCompositionMode: boolean;
+
+  constructor(options: ?{ isCompositionMode: boolean }) {
```
supersonicclay: @georgios-uber actually created a CompositeMode. It may need tweaks, but I wonder if TransformMode can use CompositeMode inside itself so we don't need to have these smaller modes aware if they're being composed.

modules/edit-modes/src/lib/rotate-mode.js
```diff
@@ -1,84 +1,195 @@
...
+  getIntermediatePosition
+} from './geojson-edit-mode.js';
 import { ImmutableFeatureCollection } from './immutable-feature-collection.js';
 
 export class RotateMode extends BaseGeoJsonEditMode {
-  _isRotatable: boolean;
-  _geometryBeingRotated: ?FeatureCollection;
+  _selectedEditHandle: ?EditHandleFeature;
+  _selectedGeometry: ?FeatureCollection;
+  _previousMouseCoords: ?Position;
```
supersonicclay: You can just use `props.lastPointerMoveEvent`

modules/edit-modes/src/lib/scale-mode.js
```diff
@@ -1,101 +1,223 @@
...
-  _isScalable: boolean;
-  _geometryBeingScaled: ?FeatureCollection;
+  _selectedGeometry: ?FeatureCollection;
+  _selectedEditHandle: ?EditHandleFeature;
+  _cornerGuidePoints: Array<EditHandleFeature>;
+  _previousMouseCoords: ?Position;
```
supersonicclay: You can just use `props.lastPointerMoveEvent`

modules/edit-modes/src/lib/rotate-mode.js
```diff
@@ -1,84 +1,195 @@
...
+      const { type } = getGeom(features[0]);
+      return type === 'Point';
+    }
+    return false;
+  };
+
+  getIsRotating = () => this._isRotating;
+
+  getGuides(props: ModeProps<FeatureCollection>) {
+    this._selectedGeometry = this.getSelectedFeaturesAsFeatureCollection(props);
```
supersonicclay: I think we should tweak this so that we capture the selected geometry before rotation (in `handleStartDragging` as it was before). This way, rounding errors don't accumulate. This happens when you rotate something, then rotate the rotated geometry, then repeat over and over on mouse move.

See this demo of the rounding errors accumulating. Notice how I start dragging and stop dragging at the top, yet the geometry is not close to its original rotation.
rotate-bug

modules/edit-modes/src/types.js
```diff
@@ -85,15 +85,22 @@ export type Tooltip = {
...
 
 export type EditHandleFeature = FeatureWithProps<
   Point,
   {
     guideType: 'editHandle',
     editHandleType: EditHandleType,
     featureIndex: number,
-    positionIndexes: number[]
+    positionIndexes: number[],
+    index?: number
```
supersonicclay: Looks like this is just used for scale mode. What do you think of just reusing `positionIndexes` instead of adding another property?

Also, noticed that `positionIndexes` should be nullable.

```
guideType: 'editHandle',
  editHandleType: EditHandleType,
  featureIndex: number,
  positionIndexes?: number[]
```

supersonicclay:
Thanks for implementing this new mode Austin, and for improving the existing modes. Great addition!

Issue #330, opened 2020-02-06
react-map-gl-draw double click in EditorModes.DRAW_POLYGON does not finalize the polygon


vmary2014:
Hi
Fabulous library thanks for the efforts going into this, really appreciate it.
I think there might be an issue when the Editor is set to **EditorModes.DRAW_POLYGON**
When I draw a polygon on the map and double click I was expecting it to finalize the polygon, as per the MapBox example [here](https://docs.mapbox.com/mapbox-gl-js/example/mapbox-gl-draw/).
However it doesn't and instead you are forced to close the polygon by clicking last on the first point/handle you drew on the map.
Is there a way to get it to do that ? I could not find any doc about this.
Thanks.
Vincent
#### To Do List 
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

xintongxia:
@vmary2014 current `react-map-gl-draw` doesn't capture the double click event. I'll evaluate the complexity of this feature.

christensenmichael0:
@vmary2014 @xintongxia will this update also apply when terminating a polyline with a double click?

xintongxia:
avmary2014 @christensenmichael0 double click to finalize a polygon is added in  v0.18.3

vmary2014:
> avmary2014 @christensenmichael0 double click to finalize a polygon is added in v0.18.3

Thanks so much for doing this ! Much appreciated.

PR #341, opened 2020-02-26
react-map-gl-draw: Closing polygon drawing on dblclick

ilyabo:
This PR addresses the issue #330

We did some user testing for polygon drawing and realized that very few users discovered that they have to connect the last and the first points to close a polygon. Double click is a standard way of doing this. Hence, it's very important to support it.

cc @xintongxia

### DIFF
--- a/modules/react-map-gl-draw/src/edit-modes/base-mode.js
+++ b/modules/react-map-gl-draw/src/edit-modes/base-mode.js
@@ -24,6 +24,8 @@ export default class BaseMode {
 
   handleClick(event: ClickEvent, props: ModeProps<FeatureCollection>) {}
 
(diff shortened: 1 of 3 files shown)


ilyabo:
The CI build is failing because of the missing `FOSSA_API_KEY`. How can I fix this?

georgios-uber:
> The CI build is failing because of the missing `FOSSA_API_KEY`. How can I fix this?

Seems like the build is broken, just ignore for now, I'm trying to fix.

georgios-uber:
@ilyabo  can you push again? I fixed travis ci.

xintongxia:
thanks for your contribution. The current way that react-map-gl-draw handled closing a polygon was from an internal use case, and I think double click to close a polygon also makes a lot sense.

PR #342, opened 2020-02-27
react-map-gl-draw: Starting to draw rect on mousedown not click

ilyabo:
Currently, to start drawing a rectangle you need to click and release the mouse. We did some user testing and realized that many users were confused by that. They expected the drawing to be initiated on mousedown, not after completing a click. This PR introduces the change.

This is also how rectangle drawing works on http://geojson.io/

It would be good in addition to still support map panning in the rect drawing mode. Many GIS tools (e.g. QGIS) let you pan the map while holding space bar while in selection by drawing mode. I tried to adding key event handlers for this, but for some reason that didn't work (they were never called):
image

I realize that you might not necessarily agree with the proposed change. But I'd like at least to discuss the possibility with you, because it's a kind of a blocker for us.

### DIFF
--- a/modules/react-map-gl-draw/src/edit-modes/base-mode.js
+++ b/modules/react-map-gl-draw/src/edit-modes/base-mode.js
@@ -26,6 +26,8 @@ export default class BaseMode {
 
   handlePointerMove(event: PointerMoveEvent, props: ModeProps<FeatureCollection>) {}
 
+  handlePan(event: ClickEvent, props: ModeProps<FeatureCollection>) {}
+
   handleStartDragging(event: StartDraggingEvent, props: ModeProps<FeatureCollection>) {}
 
   handleStopDragging(event: StopDraggingEvent, props: ModeProps<FeatureCollection>) {}
--- a/modules/react-map-gl-draw/src/edit-modes/draw-rectangle-mode.js
+++ b/modules/react-map-gl-draw/src/edit-modes/draw-rectangle-mode.js
@@ -5,7 +5,7 @@ import type {
   ClickEvent,
   FeatureCollection,
   StopDraggingEvent
-} from '@ne bula.gl/edit-modes';
+} from '@nebula.gl/edit-modes';
 import uuid from 'uuid/v1';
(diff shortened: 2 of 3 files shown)


georgios-uber:
@ilyabo please fix formatting using `yarn pretty` or `npm run pretty`.

ilyabo:
> @ilyabo please fix formatting using `yarn pretty` or `npm run pretty`.

@georgios-uber  Done!

xintongxia:
@ilyabo 

Thanks for your contribution. To unblock, I can land your changes and publish a new version. Also will create an issue to track the eventsf problems you mentioned.

The rectangle behavior is from an internal use, so probably still needs to be kept.

Users always like customization and flexibilities, so I would like to support customized modes in near future.

Proposal will be changing the Editor's options `mode` type to be a `ModeHandler` object and expose all the existing modes. And the customized mode needs to follow the `BaseMode`

- `mode` (Object, Optional)
  - default to `ReadOnlyMode`

ilyabo:
Thanks @xintongxia! If you could merge it it would be awesome. And I am not sure anymore whether the space-dragging behavior I mentioned is necessary at all (e.g. http://geojson.io/ don't have it), so for us it's good as it is now without it. We are actually now using our [own fork with the two fixes applied](https://github.com/teralytics/nebula.gl) and publish it in our internal npm repo. But it would be better, of course, to go back to the main library version.

supersonicclay:
@xintongxia does `react-map-gl-draw` now work with `DrawRectangleMode`? If so, users should be able to use `modeConfig.dragToDraw` to achieve this now.
https://github.com/uber/nebula.gl/blob/master/docs/api-reference/modes/overview.md#drawrectanglemode

supersonicclay:
@ilyabo We integrated `react-map-gl-draw` with the rest of the more powerful editing modes from the `@nebula.gl/edit-modes` package.

I think you can achieve this now with using `DrawRectangleMode` and specifying `modeConfig.dragToDraw=true`. I'm going to close this PR, but let us know if you have any issues.

PR #349, opened 2020-03-13
Add new editor module and example app

supersonicclay:
* Adds a new `@nebula.gl/editor` module
* Adds a super-basic `Toolbox` component just to show how props and callbacks can flow
* Adds a new `editor` example so that we can experiment.


examples/editor/yarn.lock
```diff
@@ -0,0 +1,5996 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
georgios-uber: We don't commit the lockfiles inside examples.

yarn.lock
```diff
@@ -8356,6 +8356,15 @@ react@^16.3.0:
     prop-types "^15.6.2"
     scheduler "^0.13.6"
 
+react@^16.8.0:
+  version "16.13.0"
+  resolved "https://unpm.uberinternal.com/react/-/react-16.13.0.tgz#d046eabcdf64e457bbeed1e792e235e1b9934cf7"
```
georgios-uber: Do not use uberinternal! Please fix.

examples/editor/yarn.lock
supersonicclay: We have been for the other ones. Let's tackle removing them if it makes sense in another PR.

PR #360, opened 2020-03-23
react-map-gl-draw: customized modes

xintongxia:
For https://github.com/uber/nebula.gl/pull/342


supersonicclay:
I think we need to refactor to use @nebula.gl/edit-modes instead of maintaining all this forked code.

xintongxia:
@ilyabo

PR #362, opened 2020-03-23
Draw polygon by dragging mode

### LINKED ISSUE #172: Freehand Draw Mode
#### Feature request
Free Hand Draw Mode
#### Background 
The ability to draw on the map without having to click to add a point. The line would trace and be added based on where your mouse moved on top of the map. 

Thank you!

keller-mark:
Fixes #172

### DIFF
--- a/modules/edit-modes/src/lib/draw-polygon-by-dragging-mode.js
+++ b/modules/edit-modes/src/lib/draw-polygon-by-dragging-mode.js
@@ -60,8 +60,6 @@ export class DrawPolygonByDraggingMode extends DrawPolygonMode {
 
     if (!clickedEditHandle) {
       // Don't add another point right next to an existing one.
-      // Also ensure that dragging has not yet ended,
-      // since the function call may have been delayed due to throttling.
       this.addClickSequence(event);
     }
   }
--- a/modules/edit-modes/src/lib/draw-polygon-by-dragging-mode.js
+++ b/modules/edit-modes/src/lib/draw-polygon-by-dragging-mode.js
@@ -16,7 +16,6 @@ type DraggingHandler = (event: DraggingEvent, props: ModeProps<FeatureCollection
 
 export class DrawPolygonByDraggingMode extends DrawPolygonMode {
   handleDraggingThrottled: ?DraggingHandler = null;
-  handleDraggingAux: DraggingHandler;
 
   handleClick(event: ClickEvent, props: ModeProps<FeatureCollection>) {
     // No-op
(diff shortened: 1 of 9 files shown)


modules/edit-modes/src/lib/draw-polygon-by-dragging-mode.js
```diff
@@ -0,0 +1,46 @@
+// @flow
+
+import type { ClickEvent, ModeProps } from '../types.js';
+import type { Polygon, FeatureCollection } from '../geojson-types.js';
+import { getPickedEditHandle } from '../utils.js';
+import { DrawPolygonMode } from './draw-polygon-mode';
+
+export class DrawPolygonByDraggingMode extends DrawPolygonMode {
```
supersonicclay: I'm super-thrilled to see the `DrawPolygonMode` flexible enough to be reused here!

modules/edit-modes/src/lib/draw-polygon-by-dragging-mode.js
```diff
@@ -0,0 +1,46 @@
...
+        props.onEdit(editAction);
+      }
+    }
+  }
+
+  handleDragging(event: ClickEvent) {
+    const { picks } = event;
+    const clickedEditHandle = getPickedEditHandle(picks);
+
+    if (!clickedEditHandle) {
```
supersonicclay: You may want to have some additional throttling logic here so you don't add thousands f points unnecessarily. Currently, points will be added as fast as the browser delivers pointer move events (and it doesn't overlap an existing point). But that may be more fidelity than some users need.

It could be a configuration that users can put in `modeConfig`. Something like `modeConfig.throttleMs` whose value is the number of milliseconds to wait after adding a point before adding another point. https://www.npmjs.com/package/lodash.throttle can be used to accomplish the throttling.

modules/edit-modes/src/lib/draw-polygon-by-dragging-mode.js
```diff
@@ -0,0 +1,75 @@
...
+    }
+  }
+
+  handleDraggingAux(event: DraggingEvent, props: ModeProps<FeatureCollection>) {
+    const { picks } = event;
+    const clickedEditHandle = getPickedEditHandle(picks);
+
+    if (!clickedEditHandle) {
+      // Don't add another point right next to an existing one.
+      // Also ensure that dragging has not yet ended,
```
supersonicclay: looks like a stale comment

supersonicclay:
Thanks for this @keller-mark !

keller-mark:
No problem, thanks for your suggestions!

supersonicclay:
Published 0.17.4

PR #379, opened 2020-04-08
De-structure coordinates prop to avoid passing to the `div`

NameFILIP:
Otherwise, it generates the following snapshot:

```
         <div
     +     coordinates={
     +       Array [
     +         200,
     +         200,
     +       ]
     +     }
           style={
```

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -51,7 +51,9 @@ module.exports = {
     'func-style': 'off',
 
     /* Use the 'query-string' module instead */
-    'no-restricted-imports': ['error', 'querystring']
+    'no-restricted-imports': ['error', 'querystring'],
+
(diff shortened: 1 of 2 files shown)


.eslintrc.js
```diff
@@ -40,7 +40,9 @@ module.exports = {
     'func-style': 'off',
 
     /* Use the 'query-string' module instead */
-    'no-restricted-imports': ['error', 'querystring']
+    'no-restricted-imports': ['error', 'querystring'],
+    
+    'no-unused-vars': ['error', { 'ignoreRestSiblings': true }]
```
NameFILIP, line 45: See https://eslint.org/docs/rules/no-unused-vars#ignorerestsiblings

PR #381, opened 2020-04-09
docs: add note about HtmlOverlayItem must be direct children of Overlay

wuweiweiwu:
question: should `HtmlOverlayItem` be able to be nested inside children components inside an `Overlay`?


georgios-uber:
Thanks.

PR #384, opened 2020-04-09
edit-modes: fix dependencies and resolving dependencies in local mode

xintongxia:
for https://github.com/uber/react-map-gl/issues/1062

`edit-modes` should be independent of `@deck.gl/core`

Alternatively we could move `memoize` to `@math.gl/core`


examples/webpack.config.local.js
```diff
@@ -11,12 +11,12 @@ const resolve = require('path').resolve;
 const webpack = require('webpack');
 
 const LIB_DIR = resolve(__dirname, '..');
-const MAIN_SRC_DIR = resolve(LIB_DIR, './modules/main/src');
-const EDIT_MODES_SRC_DIR = resolve(LIB_DIR, './modules/edit-modes/src');
-const LAYERS_SRC_DIR = resolve(LIB_DIR, './modules/layers/src');
-const OVERLAYS_SRC_DIR = resolve(LIB_DIR, './modules/overlays/src');
-const EDITOR_SRC_DIR = resolve(LIB_DIR, './modules/editor/src');
-const REACT_EDITOR_LITE_SRC_DIR = resolve(LIB_DIR, './modules/react-map-gl-draw/src');
+const MAIN_SRC_DIR = resolve(LIB_DIR, './modules/main');
```
supersonicclay, line 14: Can you see if adding the react-map-gl example to lerna.json makes this still necessary?

PR #386, opened 2020-04-10
Document the data property passed to onUpdate listeners

tschaub:
It looks like the object passed to the `onUpdate` function has a `data` property instead of `features`.

https://github.com/uber/nebula.gl/blob/20613f8697f1153d52d6af811af4ef298a231fc3/modules/react-map-gl-draw/src/mode-handler.tsx#L223-L227


xintongxia:
thank you!

PR #389, opened 2020-04-17
Fix double-click to complete

### LINKED ISSUE #387: Double-click does not complete polygon or line string
#### Describe the bug
When drawing a line or polygon, double-clicking should complete the polygon or line. 
#### To Do List 
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

Apr-16-2020 08-47-27

supersonicclay:
Fixes #387

### DIFF
--- a/modules/layers/src/layers/editable-geojson-layer.ts
+++ b/modules/layers/src/layers/editable-geojson-layer.ts
@@ -333,6 +333,10 @@ export default class EditableGeoJsonLayer extends EditableLayer {
       lastPointerMoveEvent: this.state.lastPointerMoveEvent,
       cursor: this.state.cursor,
       onEdit: (editAction: EditAction<FeatureCollection>) => {
+        // call empty setState in order to trigger a re-render
+        // This supports double-click where we need to ensure that there's a re-render between the two clicks
+        // even though the data wasn't changed, just the internal tentative feature.
+        this.setState({});
         props.onEdit(editAction);
       },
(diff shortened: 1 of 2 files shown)


modules/layers/src/layers/editable-geojson-layer.ts
```diff
@@ -333,6 +333,10 @@ export default class EditableGeoJsonLayer extends EditableLayer {
       lastPointerMoveEvent: this.state.lastPointerMoveEvent,
       cursor: this.state.cursor,
       onEdit: (editAction: EditAction<FeatureCollection>) => {
+        // call empty setState in order to trigger a re-render
+        // This supports double-click where we need to ensure that there's a re-render between the two clicks
+        // even though the data wasn't changed, just the internal tentative feature.
+        this.setState({});
```
xintongxia: I feel kind of not quite clean to force a re-render in every `onEdit` call, as this is not just for handling double click

How does `editable-geojson-layer.ts` handle double click? why the first click didn't trigger a re-render?

supersonicclay: The first click calls `onEdit` callback and updates the tentative feature object. Neither of those change the props that are handed into `EditableGeoJsonLayer` on the next react render cycle. So, deck.gl short-circuits since the props are the same and doesn't re-render.

I updated it to call `setNeedsUpdate` instead of `setState`, which I think is more appropriate.

PR #410, opened 2020-05-30
Remove double click handlers from nebula layers and edit-modes

austintang:
With the current implementation, all layers that extend `EditableLayer` add event handlers for the `dblclick` event even when `EditableLayer.onDoubleClick()` isn't overridden. The problem is that deck.gl performs click debouncing when `dblclick` handler is provided, making clicks feel laggy and in some cases, drop clicks altogether.

The following is how `EditableLayers` currently behave. Notice the delay in the layer's response after clicking on features:
debounced_clicks

Here is how the layer behaves after applying this diff:
fast_click

As a side note, if a single layer adds a `dblclick` handler, all deck layers will experience this debouncing effect. Because of this, I do not recommend that any layers add `dblclick` handlers.

### DIFF
--- a/examples/advanced/src/example.tsx
+++ b/examples/advanced/src/example.tsx
@@ -1021,8 +1021,7 @@ export default class Example extends React.Component<
               id: 'basemap',
               controller: {
                 type: MapController,
-                // @ts-ignore
-                doubleClickZoom: this.state.mode === 'view' && !this.state.selectionTool,
+                doubleClickZoom: false,
               },
             }),
           ]}
--- a/modules/react-map-gl-draw/src/mode-handler.tsx
+++ b/modules/react-map-gl-draw/src/mode-handler.tsx
@@ -54,7 +54,6 @@ export default class ModeHandler extends React.PureComponent<EditorProps, Editor
 
     this._events = {
       anyclick: (evt) => this._onEvent(this._onClick, evt, true),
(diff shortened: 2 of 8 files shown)


examples/advanced/src/example.tsx
```diff
@@ -1021,8 +1021,7 @@ export default class Example extends React.Component<
               id: 'basemap',
               controller: {
                 type: MapController,
-                // @ts-ignore
-                doubleClickZoom: this.state.mode === 'view' && !this.state.selectionTool,
```
austintang, line 1025: I feel like disabling double click to zoom will make nebula feel more snappy and responsive. In my opinion, the benefits of this change outweigh the utility that double click to zoom provides.

modules/react-map-gl-draw/src/mode-handler.tsx
```diff
@@ -308,12 +307,6 @@ export default class ModeHandler extends React.PureComponent<EditorProps, Editor
     this._modeHandler.handleClick(event, modeProps);
   };
 
-  _onDblclick = (event: BaseEvent) => {
```
xintongxia, line 311: react-map-gl-draw is not using deck.gl as renderer. so this should be kept.

austintang: Got it. I'll revert my changes in the `react-map-gl-draw` module.

PR #434, opened 2020-07-07
Change the geometry of polygons and rectangles ring directions. Outer rings use CCW and inner rings follow CW. 

GoCardsCai:
- This change to use rewind to reshape outer rings to CCW and inner rings to CW.  Here is the details about rewind: https://www.npmjs.com/package/@turf/rewind.

### DIFF
--- a/modules/edit-modes/src/lib/draw-rectangle-mode.ts
+++ b/modules/edit-modes/src/lib/draw-rectangle-mode.ts
@@ -1,10 +1,11 @@
 import bboxPolygon from '@turf/bbox-polygon';
+import rewind from '@turf/rewind';
 import { Position, Polygon, FeatureOf } from '../geojson-types';
 import { TwoClickPolygonMode } from './two-click-polygon-mode';
 
 export class DrawRectangleMode extends TwoClickPolygonMode {
   getTwoClickPolygon(coord1: Position, coord2: Position, modeConfig: any): FeatureOf<Polygon> {
-    const rectangle = bboxPolygon([coord1[0], coord1[1], coord2[0], coord2[1]]);
+    const rectangle = rewind(bboxPolygon([coord1[0], coord1[1], coord2[0], coord2[1]]));
     rectangle.properties = rectangle.properties || {};
     rectangle.properties.shape = 'Rectangle';
 
--- a/modules/edit-modes/src/lib/draw-rectangle-mode.ts
+++ b/modules/edit-modes/src/lib/draw-rectangle-mode.ts
@@ -1,15 +1,14 @@
 import bboxPolygon from '@turf/bbox-polygon';
-import rewind from '@turf/rewind';
 import { Position, Polygon, FeatureOf } from '../geojson-types';
 import { TwoClickPolygonMode } from './two-click-polygon-mode';
 
 export class DrawRectangleMode extends TwoClickPolygonMode {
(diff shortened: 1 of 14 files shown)


austintang:
Can the title of this diff be changed? This change impacts more than just rectangles. Also the `@turf/rewind` utility makes outer rings CCW and inner rings CW. Perhaps the diff title can mention more than just CCW.

GoCardsCai:
> Can the title of this diff be changed? This change impacts more than just rectangles. Also the `@turf/rewind` utility makes outer rings CCW and inner rings CW. Perhaps the diff title can mention more than just CCW.

Thanks, changed.

modules/edit-modes/src/lib/draw-rectangle-mode.ts
```diff
@@ -1,10 +1,11 @@
 import bboxPolygon from '@turf/bbox-polygon';
+import rewind from '@turf/rewind';
 import { Position, Polygon, FeatureOf } from '../geojson-types';
 import { TwoClickPolygonMode } from './two-click-polygon-mode';
 
 export class DrawRectangleMode extends TwoClickPolygonMode {
   getTwoClickPolygon(coord1: Position, coord2: Position, modeConfig: any): FeatureOf<Polygon> {
-    const rectangle = bboxPolygon([coord1[0], coord1[1], coord2[0], coord2[1]]);
+    const rectangle = rewind(bboxPolygon([coord1[0], coord1[1], coord2[0], coord2[1]]));
```
supersonicclay, line 8: Can you make this more generic by adding the rewind call in `TwoClickPolygonMode` where it calls `getTwoClickPolygon`?

modules/edit-modes/src/lib/draw-rectangle-using-three-points-mode.ts
```diff
@@ -16,7 +17,7 @@ export class DrawRectangleUsingThreePointsMode extends ThreeClickPolygonMode {
       properties: {
         shape: 'Rectangle',
       },
-      geometry: {
+      geometry: rewind({
```
supersonicclay, line 20: Can you make this more generic by adding the rewind call in `ThreeClickPolygonMode` where it calls `getThreeClickPolygon`?

supersonicclay: I should be more explicit. Can you make this change in `ThreeClickPolygonMode.handleClick` instead? 

this:
```
const editAction = this.getAddFeatureOrBooleanPolygonAction(tentativeFeature.geometry, props);
```

can become
```
const editAction = this.getAddFeatureOrBooleanPolygonAction(rewind(tentativeFeature.geometry), props);
```

Then every mode that extends `ThreeClickPolygonMode` won't need to change.

modules/edit-modes/src/lib/draw-rectangle-mode.ts
supersonicclay: I should be more explicit. Can you make this change in `TwoClickPolygonMode.handleClick` instead? 

this:
```
const editAction = this.getAddFeatureOrBooleanPolygonAction(feature, props);
```

can become
```
const editAction = this.getAddFeatureOrBooleanPolygonAction(rewind(feature), props);
```

Then every mode that extends `TwoClickPolygonMode` won't need to change.

PR #437, opened 2020-07-09
Fix getScreenCoords for Edge Browser

athomann:
Microsoft Edge getBoundingClientRect doesn't support `.x` or `.y`.  In Edge, I would receive an `invalid pixel coordinate` error for mouse events.


supersonicclay:
Thanks @athomann!

PR #444, opened 2020-07-17
Expose edit-modes utils

### LINKED ISSUE #421: Request to export utils in index.ts
#### Feature request
Kindly export utils for downstream use in index.ts the way types and geojson_types are exported.

#### Background 
https://github.com/uber/nebula.gl/blob/master/modules/edit-modes/src/index.ts
I am trying to play with some of the edit modes and it'll be great if the utils can be exported too.

#### To Do List 
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Doc update
  - [ ] Test

wootencl:
Exposes edit-modes' utils as `utils` from `index.ts`

Resolves #421.

### DIFF
--- a/modules/edit-modes/src/index.ts
+++ b/modules/edit-modes/src/index.ts
@@ -81,4 +81,5 @@ export type {
 } from './geojson-types';
 
 // Utils
-export * as utils from './utils';
+import * as utils from './utils';
+export { utils };
--- a/modules/edit-modes/src/index.ts
+++ b/modules/edit-modes/src/index.ts
@@ -79,3 +79,6 @@ export type {
   FeatureCollection,
   AnyGeoJson,
 } from './geojson-types';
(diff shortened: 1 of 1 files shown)


modules/edit-modes/src/index.ts
```diff
@@ -79,3 +79,6 @@ export type {
   FeatureCollection,
   AnyGeoJson,
 } from './geojson-types';
+
+// Utils
+export * as utils from './utils';
```
supersonicclay, line 84: Looks like this syntax isn't quite right with the version of TS we're using. The build is failing.

wootencl: Would have helped if I actually tried to build 🤦‍♂️. Pushed fix up/should be resolved now @supersonicclay.

PR #445, opened 2020-07-20
react-map-gl-draw: Handle lastPointerMoveEvent null case

ijosc:
`lastPointerMoveEvent` is initialised as `null`. If the user does not interact with the map, i.e. no `PointerMoveEvent` has been registered, a call to `_getCursorEditHandle()` will fail here: `const { isDragging, picks } = event;`, as event is still `null`.

Our use case is that we want to be able to set the `selectedFeatureIndex` from outside the map component, without requiring the user to have interacted with the map.


daniel-j-h:
:wave: @xintongxia @kikonejacob - any chance we can get your :eyes: on this one? :hugs: 

I'm running into this issue and the fix is quick and does the job :rocket: 

Any thoughts? :thought_balloon:

PR #454, opened 2020-07-30
Better handle null/undefined mode prop

### LINKED ISSUE #452: Can't import modes from @deck.gl/layers
#### Describe the bug
```
import {DrawPolygonMode} from '@nebula.gl/layers'
```

#### Actual Result
Can't call getGuides on undefined

If you change to 
```
import {DrawPolygonMode} from '@nebula.gl/edit-modes'
```
then it works

#### Expected Result
Works

#### Reproduce Steps
https://codesandbox.io/s/deckgl-and-nebulagl-editablegeojsonlayer-no-react-p9yrs

#### To Do List 
  - [x] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

supersonicclay:
Fixes #452

### DIFF
--- a/modules/layers/src/layers/editable-geojson-layer.ts
+++ b/modules/layers/src/layers/editable-geojson-layer.ts
@@ -32,6 +32,7 @@ import {
   DraggingEvent,
   PointerMoveEvent,
   GeoJsonEditModeType,
+  GeoJsonEditModeConstructor,
   FeatureCollection,
 } from '@nebula.gl/edit-modes';
 
@@ -98,7 +99,7 @@ function getEditHandleRadius(handle) {
 }
 
 const defaultProps = {
-  mode: DrawPolygonMode,
+  mode: DEFAULT_EDIT_MODE,
 
   // Edit and interaction events
   onEdit: () => {},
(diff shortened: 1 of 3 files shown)


supersonicclay:
cc @andybak

modules/layers/src/layers/editable-geojson-layer.ts
```diff
@@ -285,11 +286,17 @@ export default class EditableGeoJsonLayer extends EditableLayer {
...
     super.updateState({ props, changeFlags });
 
-    let mode = this.state.mode;
     if (changeFlags.propsOrDataChanged) {
-      if (props.mode !== oldProps.mode) {
-        if (typeof props.mode === 'string') {
-          // Lookup the mode based on its name (for legacy purposes)
+      const modePropChanged = Object.keys(oldProps).length === 0 || props.mode !== oldProps.mode;
+      if (modePropChanged) {
+        let mode = this.state.mode;
```
xintongxia, line 292: this isn't used, `mode` is set based on `props.mode` later.

modules/layers/src/layers/editable-geojson-layer.ts
```diff
@@ -299,15 +306,10 @@ export default class EditableGeoJsonLayer extends EditableLayer {
           mode = props.mode;
```
xintongxia, line 306: if `mode` isnt `string` or `function`, what it would be?

supersonicclay: An instance of an EditMode.

PR #458, opened 2020-08-06
feat: Convert class properties to overridable methods.

faeron:
The usage of class properties (via babel-plugin) prevents overrides for that methods when extending. Now EditMode can be extended and all methods can be customized.


supersonicclay:
@xintongxia can you give it a look and merge?

PR #479, opened 2020-09-21
Create EditableH3Layer && Refactor Edit Modes to support updateTentativeFeature edit type

mikecolonna:
**6e18657: Refactor Edit Modes to support updateTentativeFeature edit type**:

This change refactors the following `GeoJsonEditMode`'s to support the new `updateTentativeFeature` edit type:

- Draw90DegreePolygonMode
- DrawPolygonMode (and all extended modes)
- GeoJsonEditMode
- TwoClickPolygonMode (and all extended modes)
- ThreeClickPolygonMode (and all extended modes)

`updateTentativeFeature` is a new edit type meant to be fired whenever a polygonal shape is near completion, and the user continues to drag the cursor mid-editing. It fires on pointer move. The tentative feature created is different for each mode. Tested for all of the above modes.

This edit type will be used to update a cluster of hexes on pointer move in an `EditableH3Layer`.

**85fc681: Create EditableH3Layer**:

This creates a new `EditableH3Layer`. This layer uses the `EditableGeoJsonLayer` to support editing and display of H3 hexagons via the Deck.gl `H3ClusterLayer`. This layer supports the `union`, `difference`, and `intersection` draw modes.
Example usage: https://imgur.com/2JGBbZS


modules/layers/src/layers/editable-h3-layer.ts
```diff
@@ -0,0 +1,152 @@
...
+  stroked: true,
+  lineWidthScale: 1,
+  lineWidthMinPixels: 1,
+  lineWidthMaxPixels: Number.MAX_SAFE_INTEGER,
+  lineWidthUnits: 'pixels',
+  getHexagons: (d) => d.hexIds,
+  resolution: DEFAULT_H3_RESOLUTION,
+};
+
+export default class EditableH3Layer extends EditableLayer {
```
supersonicclay, line 31: I think it'd be great to have an example app demonstrating this layer as part of this PR.

modules/layers/src/layers/editable-h3-layer.ts
```diff
@@ -0,0 +1,152 @@
...
+                  this.setState({ tentativeHexagonIDs: hexIDs });
+                }
+                break;
+              case 'addFeature':
+                // once the shape is finished drawing, calculate final hexagons
+                let finalHexagonIDs;
+                const committedHexagonIDs = new Set(this.getHexIdsFromH3Data());
+                const tentativeHexagonIDs = new Set(this.state.tentativeHexagonIDs);
+
+                switch (this.props.modeConfig.booleanOperation) {
```
supersonicclay, line 79: allow null `modeConfig`
```
const {modeConfig} = this.props;
switch (modeConfig && modeConfig.booleanOperation) {
...
```

modules/layers/src/layers/editable-h3-layer.ts
```diff
@@ -0,0 +1,152 @@
...
+
+const defaultProps = {
+  mode: DEFAULT_EDIT_MODE,
+
+  // EditableGeoJsonLayer
+  geoJsonData: {},
+  ...EditableGeoJsonLayer.defaultProps,
+
+  // h3 layer
+  h3Data: {},
```
supersonicclay, line 20: to make it consistent with other deck.gl layers, can we use the name `data` instead of `h3Data` for the prop name?

supersonicclay: also, default should be empty array `[]` instead of empty object `{}` I believe.

modules/layers/src/layers/editable-h3-layer.ts
```diff
@@ -0,0 +1,152 @@
...
+      ),
+    ];
+    return layers;
+  }
+
+  // because h3Data is an array of hexagon data, we take the cumulative hexIds from all entries
+  getHexIdsFromH3Data() {
+    let cumulativeHexIds = [];
+    this.props.h3Data.forEach(
+      (h3Data) => (cumulativeHexIds = cumulativeHexIds.concat(h3Data.hexIds))
```
supersonicclay, line 139: Seems you should use `this.props.getHexagons` instead of hardcoding here.

modules/layers/src/layers/editable-h3-layer.ts
```diff
@@ -0,0 +1,152 @@
...
+                this.props.onEdit({ updatedData: finalHexagonIDs });
+                break;
+              default:
+                break;
+            }
+          },
+        })
+      ),
+
+      new H3ClusterLayer(
```
supersonicclay, line 112: you may want to render the tentative hexagons last, so they render on top. See https://codesandbox.io/s/nebulagl-h3-hexagon-editing-xgntv?file=/index.js

modules/layers/src/layers/editable-h3-layer.ts
mikecolonna: That's fair. This behavior should be left up to the caller.

modules/layers/src/layers/editable-h3-layer.ts
mikecolonna: Yup, makes sense.

modules/layers/src/layers/editable-h3-layer.ts
mikecolonna: Good call. I'll put one together.

supersonicclay: You can use this as a starting point:
https://codesandbox.io/s/nebulagl-h3-hexagon-editing-xgntv?file=/index.js

supersonicclay, line 31: I think we can make this closer to existing APIs by doing the following:
1. rename to `EditableH3ClusterLayer`.
2. take a `selectedIndexes` prop which is an array of indexes into the `data` array. It's ok if you only support single selection today (`selectedIndexes.length === 1`), but taking an array of indexes future-proofs the API a bit. And it will support operations in the future that may operate on multiple selection (e.g. translate, scale, etc).

```
new EditableH3ClusterLayer({
  data: [{h3Cells: [...]}], // valid
  data: [{h3Cells: [...]}, {h3Cells: [...]}], // valid? (based on how much time you have)
  selectedIndexes: [0], // valid
  selectedIndexes: [1], // valid? (based on how much time you have)
  selectedIndexes: [0, 1], // invalid
  selectedIndexes: [], // invalid? (based on how much time you have)
  booleanOperation: null, // valid? (based on how much time you have)
  booleanOperation: 'union', // valid
  ...
})
```

modules/layers/src/layers/editable-h3-layer.ts
```diff
@@ -0,0 +1,152 @@
...
+                      tentativeHexagonIDs.has(hexID)
+                    );
+                    break;
+                  case 'difference':
+                    finalHexagonIDs = [...committedHexagonIDs].filter(
+                      (hexID: string) => !tentativeHexagonIDs.has(hexID)
+                    );
+                    break;
+                  default:
+                    break;
```
supersonicclay, line 96: perhaps we should union by default rather than doing nothing (after the user spent their time drawing a polygon).
```
switch(...) {
  case 'union':
  default:
    // do the union logic
  case 'intersection':
    ...
}
```

PR #483, opened 2020-09-29
Create EditableH3ClusterLayer && Refactor Edit Modes to support updateTentativeFeature edit type

mikecolonna:
**Refactor Edit Modes to support updateTentativeFeature edit type**:

This change refactors the following `GeoJsonEditMode`'s to support the new `updateTentativeFeature` edit type:

- Draw90DegreePolygonMode
- DrawPolygonMode (and all extended modes)
- DrawPointMode
- GeoJsonEditMode
- TwoClickPolygonMode (and all extended modes)
- ThreeClickPolygonMode (and all extended modes)

`updateTentativeFeature` is a new edit type meant to be fired whenever a polygonal shape is near completion, and the user continues to drag the cursor mid-editing. It fires on pointer move. The tentative feature created is different for each mode. Tested for all of the above modes.

This edit type will be used to update a cluster of hexes on pointer move in an `EditableH3ClusterLayer`.

**Create EditableH3ClusterLayer**:

This creates a new `EditableH3ClusterLayer`. This layer uses the `EditableGeoJsonLayer` to support editing and display of H3 hexagons via the Deck.gl `H3ClusterLayer`. This layer supports the `union`, `difference`, and `intersection` draw modes, and supports single hexagon interactions. This layer also supports selection of different hexagon clusters.
Example usage: https://imgur.com/2JGBbZS


examples/editable-h3-cluster-layer/index.js
```diff
@@ -0,0 +1,200 @@
...
+        hexIds: updatedData,
+      };
+      setH3Clusters(newH3Clusters);
+    },
+    _subLayerProps: {
+      "tentative-hexagons": {
+        getFillColor: getTentativeFillColor(booleanOperation)
+      },
+      hexagons: {
+        getFillColor: [50, 100, 200, 230]
```
supersonicclay, line 161: may want to give selected cluster a different color:
```
getFillColor: (cluster) => {
  if (selectedIndexes.some((i) => data[i] === cluster)) {
    return SELECTED_FILL_COLOR;
  } else {
    return UNSELECTED_FILL_COLOR;
  }
}
```

modules/layers/src/layers/editable-h3-cluster-layer.ts
```diff
@@ -0,0 +1,179 @@
...
+  }
+
+  renderLayers() {
+    const layers: any = [
+      new EditableGeoJsonLayer(
+        this.getSubLayerProps({
+          id: 'editable-geojson',
+
+          mode: this.props.mode,
+          data: {
```
supersonicclay, line 61: this may cause a performance issue since deck.gl will see data as a new object on every render. May need to extract this as a const.

```
const EMPTY_FEATURE_COLLECTION = {...}

export class EditableH3ClusterLayer {
...
  new EditableGeoJsonLayer({
    ...
    data: EMPTY_FEATURE_COLLECTION
  ...
}

File: examples/editable-h3-cluster-layer/index.js [hunk 3]
```diff
@@ -0,0 +1,194 @@
...
+      },
+      hexagons: {
+        getFillColor: (cluster) => {
+          if (selectedIndexes.some((i) => h3Clusters[i] === cluster)) {
+            return SELECTED_FILL_COLOR;
+          }
+          return UNSELECTED_FILL_COLOR;
+        },
+        updateTriggers: {
+          getFillColor: selectedIndexes,
```
supersonicclay, line 154: This is the way you can trigger deck.gl to recalculate some prop (in this case `getFillColor`) base on something other than `data`.

modules/layers/src/layers/editable-h3-cluster-layer.ts
```diff
@@ -0,0 +1,180 @@
...
+                        tentativeHexagonIDs.has(hexID)
+                      );
+                      break;
+                    case 'difference':
+                      finalHexagonIDs = [...committedHexagonIDs].filter(
+                        (hexID: string) => !tentativeHexagonIDs.has(hexID)
+                      );
+                      break;
+                  }
+                  this.props.onEdit({ updatedData: finalHexagonIDs });
```
supersonicclay, line 123: To match `EditableGeoJsonLayer`, this should include all the clusters, not just the affected cluster. This way, users can simplify their `onEdit` callback to this:
```
const [data, setData] = React.useState([...]);

onEdit({updatedData}) {
  if (updatedData !== data) {
    setData(updatedData);
  }
}
```

So, basically take the logic you currently have in the example app, and put it here.

examples/editable-h3-cluster-layer/index.js
mikecolonna: Ah nice! Great to know.

modules/layers/src/layers/editable-h3-cluster-layer.ts
```diff
@@ -0,0 +1,180 @@
...
+const EMPTY_FEATURE_COLLECTION = {
+  type: 'FeatureCollection',
+  features: [],
+};
+
+const defaultProps = {
+  mode: DEFAULT_EDIT_MODE,
+
+  // EditableGeoJsonLayer
+  geoJsonData: {},
```
supersonicclay, line 20: unused?

modules/layers/src/layers/editable-h3-cluster-layer.ts
```diff
@@ -0,0 +1,201 @@
...
+                  this.setState({ tentativeHexagonIDs: [hexID] });
+                }
+                break;
+              case 'addFeature':
+                // once the shape is finished drawing, calculate final hexagons
+
+                // get selected h3 cluster to act upon
+                // if there is more than 1 selected,
+                // throw an error, since it is ambiguous how to proceed
+                if (this.props.selectedIndexes.length > 1) {
```
supersonicclay, line 102: To make this more consistent with `EditableGeoJsonLayer`, can we update the logic to this?
```
if (!modeConfig || !modeConfig.booleanOperation) {
  // add new hex cluster
} else if (this.props.selectedIndexes.length !== 1) {
  // eslint-disable-next-line no-console,no-undef
  console.warn('booleanOperation only supported for single cluster selection');
  return;
} else {
  // switch on booleanOperation and affect single selected cluster
}
this.props.onEdit(...)
```

supersonicclay (approved):

supersonicclay:
Awesome work @mikecolonna. Thanks!

PR #486, opened 2020-10-14
fix: issue with updateState when autoHighlight and pickable on

contra:
Layer doesn't seem to support calling `updateState` without oldProps because of this access: https://github.com/visgl/deck.gl/blob/1454a7cbae08b8d9b8e91e0bca1f52077e5e7918/modules/core/src/lib/layer.js#L331

Screen Shot 2020-10-14 at 1 53 22 PM

Let me know if I missed the reason why oldProps isn't being passed in - if there is a purpose behind leaving it out then this should be fixed in deck.gl to make updateState work without oldProps.

### DIFF
--- a/modules/layers/src/layers/editable-geojson-layer.ts
+++ b/modules/layers/src/layers/editable-geojson-layer.ts
@@ -286,7 +286,7 @@ export default class EditableGeoJsonLayer extends EditableLayer {
     changeFlags: any;
   }) {
     // @ts-ignore
-    super.updateState({ props, changeFlags });
+    super.updateState({ oldProps, props, changeFlags });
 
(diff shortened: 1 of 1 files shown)


contra:
Specifically the issue was introduced by this PR in deck.gl: https://github.com/visgl/deck.gl/pull/4807

Pessimistress:
https://github.com/visgl/deck.gl/issues/5028

I consider this a bug in nebula.gl. There is no reason to drop arbitrary arguments when overriding a method.

supersonicclay (approved):

supersonicclay:
Thanks @contra and @Pessimistress !

PR #522, opened 2021-02-10
Fix cursor issue when using the scale mode

StijnAmeloot:
When the scale edit mode is enabled, the bidirectional resize cursor icon (nesw-resize or nwse-resize) should only be visible in the following cases:
1) when hovering above one of the handles
2) when scaling one or more features

Currently it remains visible after hovering once above one of the handles (even when not hovering or scaling).

Before (cursor not ok)
scaling-before

After (cursor ok)
scaling-after


StijnAmeloot:
Added screenshots to clarify the issue.

kargeor (approved):

supersonicclay (approved):

PR #525, opened 2021-02-14
Add ability to resize an existing circle by clicking twice

StijnAmeloot:
By default a circle can now be resized by clicking twice along the ring.
The 'dragToDraw' option has been added to the mode config to enable dragging instead of clicking twice.


StijnAmeloot:
Fixed a small issue and added a gif which shows the two different modes.
Please let me know if there is anything that needs to be improved.

igorDykhta (commented):
1) I think dragToDraw: true should be by default
2) Not sure how the mode with dragToDraw: false should work. I activated the resizing mode, then I clicked **anywhere**, no response at this point, then the circle starts to resize with any mouse movement, then I need to click to finish resizing. I assume the resizing should only start after clicking on the small helper circle which appears on the edge of the circle.

PR #549, opened 2021-04-03
Add option to prevent overlapping lines when drawing a polygon

StijnAmeloot:
Adds the 'preventOverlappingLines' mode config option to the 'draw polygon' edit mode.
When true, adding a polygon point will be prevented if the current line overlaps any other lines on the same polygon.

Note: no feedback message is shown when it's not possible to add a polygon point.

Demo
draw-polygon-without-overlapping-lines

### DIFF
--- a/modules/edit-modes/src/lib/draw-polygon-mode.ts
+++ b/modules/edit-modes/src/lib/draw-polygon-mode.ts
@@ -5,6 +5,8 @@ import {
   GuideFeatureCollection,
   TentativeFeature,
 } from '../types';
+import lineIntersect from '@turf/line-intersect';
+import { lineString as turfLineString } from '@turf/helpers';
 import { Polygon, FeatureCollection } from '../geojson-types';
 import { getPickedEditHandle } from '../utils';
 import { GeoJsonEditMode } from './geojson-edit-mode';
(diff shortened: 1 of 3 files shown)


modules/edit-modes/src/lib/draw-polygon-mode.ts
```diff
@@ -79,14 +81,27 @@ export class DrawPolygonMode extends GeoJsonEditMode {
   handleClick(event: ClickEvent, props: ModeProps<FeatureCollection>) {
     const { picks } = event;
     const clickedEditHandle = getPickedEditHandle(picks);
+    const clickSequence = this.getClickSequence();
+
+    let overlappingLines = false;
+    if (clickSequence.length > 2 && props.modeConfig && props.modeConfig.preventOverlappingLines) {
```
supersonicclay, line 87: I like this functionality as the default and only configuration for DrawPolygonMode. I can't think of a use case where an application would want to allow overlapping lines. What do you think?

Yaswanth-Tippireddy (approved):

modules/edit-modes/src/lib/draw-polygon-mode.ts
Yaswanth-Tippireddy: @supersonicclay , this is a very useful and important functionality in our use case, we use this to draw polygons of agricultural farms, and the end users are not very tech savvy  and therefore they usually dont check if they lines are overalapping at the corners etc. This results in invalid geometries/polygons with errors . (Attaching a picture for reference). Having a feature to prevent next point incase of overlapping will help us in preventing these errors. @igorDykhta Kindly merge this commit please. 
image

StijnAmeloot: @supersonicclay For me it would make sense if it is the default. In our application we added a check after creating a polygon to determine if the shape is valid or not (basically if there are overlapping lines, we don't allow the user to save). It would be nice if we could prevent the user from creating invalid polygon in the first place. I made it opt-in to ensure that the existing functionality doesn't change, but I can change that if you want.

igorDykhta (approved):

Issue #535, opened 2021-03-08
ResizeCircleMode editType value should be distinguishable in onEdit callback


KristerL:
#### Describe the bug
Resize circle mode is missing distinguishable onEdit callback editType value. OnEdit callback gets editType props. This is really useful to toggle some kind of event after editing. 

For example, when using Modify mode to drag a point in a polygon, then after dragging the final editType has a value 'finishMovePosition'. 

Scale mode has editType "scaling" when dragging and "scaled" when dragging has finished.

Now the expectation is that resize circle has similar logic - while resizing editType is something that indicates that resizing is underway and after resize it is something like "resized", "finishedRezing, etc.

#### Actual Result
Currently resizeCircleMode returns only "unionGeometry"

#### Expected Result
When resize has finished it should say something to differentiate it from the resize in progress state.

#### Reproduce Steps
Editlayer where mode is resizeCircleMode and log the onEdit callback value.

#### Screenshot
![edit](https://user-images.githubusercontent.com/25435795/110297549-4c171780-7ffc-11eb-90ad-439cc96ed0f3.png)

Version 0.22.3
#### To Do List 
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

KristerL:
As a side note I also noticed that onHover callback on editableGeojsonLayer does not correctly work when resizing circles. Probably related to this problem

PR #558, opened 2021-04-28
added circleResizeEnd onEdit callback no notify when dragging ended. …

KristerL:
Changed circle resize onEdit callback to have unique editType during resize. Also added onEdit callback to notify when dragging has ended. Issue this resolves: [535](https://github.com/uber/nebula.gl/issues/535)


.gitignore
```diff
@@ -1,7 +1,7 @@
-**/dist/*
-**/dist-esm/*
-**/dist-es6/*
-**/dist-types/*
+#**/dist/*
```
supersonicclay, line 1: These generated files shouldn't be checked into git. Can you revert them?

supersonicclay (changes requested):

PR #587, opened 2021-05-26
Update EditableGeoJSONLayer docs and update outdated links

JulesBlm:
Hi, I noticed there were many properties and data setters in the code of EditableGeoJSONLayer that were not in the docs, properties for the the point handles and tentative features like `getTentativeFillColor`. I've added these to the documentation and I've updated all links to DeckGL's website while I was add it.


ibgreen (approved):
Thanks!

Issue #580, opened 2021-05-21
[react-map-gl-draw] EditingMode no longer working since 0.22.1


xharmagne:
#### Describe the bug
When editing a drawn polygon, any functionality to do with dragging/pointer move (move polygon, edit position of new and existing points) does not work from 0.22.1. Works on 0.21.1, downgraded our version as a workaround for now.

#### Actual Result
Can add new points on polygon but can't move any of the points or the polygon

#### Expected Result
Should be able to move a polygon and drag new and existing points on a polygon

#### Reproduce Steps
- Create a polygon in DrawPolygonMode
- Switch to EditingMode
- Try to move polygon or edit points

#### Screenshots
![drag-issue](https://user-images.githubusercontent.com/11036627/119092729-21acf600-ba52-11eb-9246-32fc4cee766c.gif)

https://codesandbox.io/s/react-map-gl-draw-example-forked-9n72e

#### To Do List 
  - [ ] Add label and assign to milestone
  - [ ] Coding
  - [ ] Test

xharmagne:
Tracked it down to this code - https://github.com/uber/nebula.gl/blob/master/modules/react-map-gl-draw/src/mode-handler.tsx#L422-L426

```
if (this._modeHandler.handleDragging) {        
        this._modeHandler.handleDragging(pointerMoveEvent, modeProps);     
} else {
        this._modeHandler.handlePointerMove(pointerMoveEvent, modeProps);
}
```

That first `if` is true for `EditingMode` because it extends `BaseMode` which has an empty implementation of `handleDragging`, even though the real work for `EditingMode` is done in `handlePointerMove`

biniyam17:
@xharmagne Thanks for reporting. I'm curious why you are importing the edit modes from "react-map-gl-draw" and not '@nebula.gl/edit-modes'? Could that be causing the issue?

xharmagne:
Hey @biniyam17 , this is a fork of the codesandbox example from the react-map-gl-draw readme under "Advanced example: multiple draw modes and editing drawn features". All I changed in my fork was the version of react-map-gl-draw.

I've debugged it and it does hit the correct edit modes (i.e. under @nebula.gl/edit-modes) on click, it's just that `handlePointerMove` which does most of the work for `EditingMode` doesn't get called because of the code in [the comment above](https://github.com/uber/nebula.gl/issues/580#issuecomment-845696809).

kjkurtz:
[...]

PR #600, opened 2021-06-07
fix(react-map-gl-draw): Fix issue 580 EditingMode drag to move

xharmagne:
Fixed issue with react-map-gl-draw EditingMode where dragging to move an edit handle or a geometry wasn't working. Issue #580 

Fixed by calling`handlePointerMove` even when not dragging.
 - this is in line with the [generic edit mode RFC](https://github.com/uber/nebula.gl/blob/master/dev-docs/RFCs/v1.0/generic-edit-mode.md) which states `handlePointerMove` should be called "regardless of whether the pointer is down, up, and whether something was picked"
 - this is also in line with how nebula gl works with deck gl - `handlePointerMove` is called along with (and not instead of) `handleDragging`

### DIFF
--- a/modules/react-map-gl-draw/src/mode-handler.tsx
+++ b/modules/react-map-gl-draw/src/mode-handler.tsx
@@ -417,13 +417,11 @@ export default class ModeHandler extends React.PureComponent<EditorProps, Editor
       cancelPan: event.sourceEvent.stopImmediatePropagation,
     };
 
+    const modeProps = this.getModeProps();
+
+    this._modeHandler.handlePointerMove(pointerMoveEvent, modeProps);
     if (this.state.didDrag) {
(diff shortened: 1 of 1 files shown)


pthomas931:
Having EditingMode back would be huge!

maxnowack (approved):

supersonicclay (approved):

ecfaria:
Hey guys, can you confirm this fix is included on v1.02? It still doesn't work for me.

igorDykhta:
@ecfaria No. The last v1.02 was released on Dec 8, 2021, but the associated fix was merged to master on Dec 21, 2021.
I'll make alpha 1.03 during the next week.

ecfaria:
Thanks, @igorDykhta!

PR #611, opened 2021-07-13
Nearest point on line

apguan:
Addressed issue: https://github.com/Turfjs/turf/issues/1440

@turf/nearest-point-on-line currently does not consistently calculate the closest snapping point. As lines/polygons get larger, the precision is reduced due to different projections that are used for calculations in turf. Since nebula uses Mercator projection, Cartesian geometry should be sufficient. Deviations caused by zoom are addressed by matching points to the pixels in the viewport. If the viewport is not provided, there is slight offset, but is still more accurate than current turf implementations

### DIFF
--- a/modules/edit-modes/src/lib/modify-mode.ts
+++ b/modules/edit-modes/src/lib/modify-mode.ts
@@ -63,7 +63,7 @@ export class ModifyMode extends GeoJsonEditMode {
           [],
           (lineString, prefix) => {
             const lineStringFeature = toLineString(lineString);
-            const candidateIntermediatePoint = this.nearestPointOnLine(
+            const candidateIntermediatePoint = this.getNearestPoint(
               // @ts-ignore
               lineStringFeature,
               referencePoint,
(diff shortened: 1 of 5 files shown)


apguan:
Before: 
before

After:
after

modules/edit-modes/src/lib/modify-mode.ts
```diff
@@ -124,8 +124,7 @@ export class ModifyMode extends GeoJsonEditMode {
         'Editing 3D point but modeConfig.viewport not provided. Falling back to 2D logic.'
       );
     }
-
-    return nearestPointOnLine(line, inPoint);
+    return nearestPointOnLine(line, inPoint, viewport);
```
supersonicclay, line 127: I know it was already an issue, but can we rename one of the `nearestPointOnLine` functions? It is weird to have a `this.nearestPointOnLine` and `nearestPointOnLine` functions in the same file. Perhaps just `this.getNearestPoint` and `nearestPointOnLine`.

supersonicclay (approved):

modules/edit-modes/src/lib/modify-mode.ts
apguan: sure, can do on this.

PR #628, opened 2021-08-23
fix "this.state is null" error

Akiyamka:
When a layer is not yet ready, trying to read its cursor causes the entire layer to fall


andrei-kucherov (approved):

andrei-kucherov (approved):

ibgreen (approved):

ibgreen:
@andrei-kucherov Trying to understand why your approval wasn't sufficient to unblock merging. Looks like you have a pending invite to this repo that was not yet accepted? I sent a new invite with "write" access.

Screen Shot 2021-08-24 at 1 41 58 PM

nekitk:
Could you please release a new version with this fix?

PR #642, opened 2021-09-17
Add option to MeasureDistanceMode to center tooltips on line

NgoKnows:
### Summary
This PR adds the ability to pass a `centerTooltipsOnLine` option to the `modeConfig`, so that rather than the tooltips being placed at the end of their respective line, they are placed at the midpoint (determined by turf).

I feel like this would be desirable behavior, because when measuring multiple distances it can be easy to mix up which tooltips belong to which distances. So placing them in the middle helps alleviate that issue.

#### `centerTooltipsOnLine === false`
before-measure

#### `centerTooltipsOnLine === true`
after-measure


NgoKnows:
I realized that these measurements are cumulative, which makes this less useful. That begs the question though if we want to allow for another option that could be used in tandem with this, to have the measurements not be cumulative.

I would be happy to add this as well, but before moving forward just want to make sure that is something that would be desirable.

igorDykhta (approved):

PR #679, opened 2022-01-11
fix: link to modes overview

aaneitchik:
There are two links on this page to modes overview section, but the first one was broken on the website, so I copied the link from the working one.

If I understand correctly, the link like this should also work on the website: `[modes overview](/docs/api-reference/modes/overview)` (without `.md`), but I checked other links in the repo, and the links with similar format don't work in the GitHub READMEs (only on the website), so I chose the one that should be working in both.


igorDykhta (approved):
Thanks!

PR #691, opened 2022-01-28
feat(modify-mode): an option to lock shape or rectangles and prevent insertion of point

igorDykhta:
- ModifyMode: modeConfig.lockRectangles option to preserve shapes of rectangular features

- [x] test
- [x] new prop logic
- [x] add node


ibgreen (approved):
add to whats-new...

PR #692, opened 2022-01-29
feat(translate-mode): translate shapes in screen space

igorDykhta:
- an option to translate shapes in screen space, as direct translation in lon/lat space distorts shapes

image


modules/edit-modes/src/utils.ts
```diff
@@ -457,3 +457,22 @@ function getEditHandlesForCoordinates(
   }
   return editHandles;
 }
+
+export function isNumeric(val: any) {
+  return !Array.isArray(val) && !isNaN(parseFloat(val)) && isFinite(val);
+}
+
+export function traverseCoords(
+  coords: any,
```
ibgreen, line 466: Can we add firmer typings for the input?

modules/edit-modes/src/utils.ts
```diff
@@ -457,3 +457,22 @@ function getEditHandlesForCoordinates(
   }
   return editHandles;
 }
+
+export function isNumeric(val: any) {
```
ibgreen, line 461:
```suggestion
export function isNumeric(val: unknown): boolean {
```

ibgreen, line 461: Add a JSDoc line for each function with a one liner explanation

ibgreen (approved):
Adding a 'screenSpace' parameter seems like a useful thing but why only for the translate mode? 

- Should we at least mention it in the docs? https://nebula.gl/docs/api-reference/modes/overview
- Without a test this will likely be fragile

belopot:
@ibgreen  @igorDykhta I had this issue with the latest version (1.0.4). Do you have any advice?

igorDykhta:
> @ibgreen @igorDykhta I had this issue with the latest version (1.0.4). Do you have any advice? https://github.com/uber/nebula.gl/assets/46432435/ddb4e0c4-bcde-491d-ac1c-7886f5019793

Have you tried to pass `screenSpace: true` for translate-mode?

belopot:
@igorDykhta Yes I tried to pass `translateInScreenSpace: true` for `TranslateMode`. However it didn't work
image

igorDykhta:
```
    modeConfig: {
      viewport,
      screenSpace: true,
    },
```

https://github.com/uber/nebula.gl/blob/b316c6348f2ddfdf819dd6128438b39c39fa6db7/modules/edit-modes/test/lib/translate-mode.test.ts#L59

PR #694, opened 2022-02-04
feat(editableGeojsonLayer): extra selection width for lines

igorDykhta:
- pickingLineWidthExtraPixels prop: Additional line width in pixels used for picking. Can be useful when `EditableGeojsonLayer` is over a deck.gl layer and precise picking is problematic, and when the usage of `pickingDepth` introduces performance issues.

### DIFF
--- a/docs/api-reference/layers/editable-geojson-layer.md
+++ b/docs/api-reference/layers/editable-geojson-layer.md
@@ -376,6 +376,12 @@ Number of pixels around the mouse cursor used for picking. This value determines
 
 Number of layers of overlapping features that will be picked. Useful in cases where features overlap.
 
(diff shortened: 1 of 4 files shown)


igorDykhta:
> * Any way to create a test case?
> * Is there a `docs/whatsnew.md` page where we should mention this.
> * Is there any developer guide about picking or FAQ doc where we should mention this trick?

1) No tests at the moment for layers in nebula.gl. Should be added, I'll open an issue.
2) I'll update changelog.md with the next release. There is no docs/whatsnew.md, I'll add it as well separately.
3) I've added the property description for EditableGeojsonLayer. No separate guide for picking.

Issue #729, opened 2022-02-23
PathMarkerLayer has incorrectly positioned arrows


NameFILIP:
#### Describe the bug
Broken path markers after the fix of https://github.com/uber/nebula.gl/pull/671
The following commit: https://github.com/uber/nebula.gl/commit/96229b0bfcd30e11ecbd39f291c4bc1c9c79b0fa is not taking care of the early exit from the loop.

Issue #669, opened 2021-12-03
TypeError: Cannot read properties of undefined (reading 'clone')


NameFILIP:
Getting the error in the `createMarkerAlongPath` function: https://github.com/uber/nebula.gl/blob/81c8372acf2032d27e43db802f65b3d8a6cf86b6/modules/layers/src/layers/path-marker-layer/create-path-markers.ts#L84

To be more specific, the error happens at `path[i + 1].clone()`.
This happens because `i` is incremented after the last cycle (because of `i++`), so the `i + 1` falls out of range anywhere after the cycle.

Simple repro:
```
const path = [1, 2, 3];

let i = 0;
for (i = 0; i < path.length - 1; i++) {
  console.log("i", i);
  console.log("path[i]", path[i]);
  console.log("path[i + 1]", path[i + 1]);
}

console.log("after the cycle");
console.log("i", i);
console.log("path[i]", path[i]);
console.log("path[i + 1]", path[i + 1]);
```

PR #730, opened 2022-02-23
 [path-marker-layer] Fix incorrectly positioned arrows

NameFILIP:
Fixes: https://github.com/uber/nebula.gl/issues/729

Contains:
* the revert of: https://github.com/uber/nebula.gl/commit/96229b0bfcd30e11ecbd39f291c4bc1c9c79b0fa
* addition of

```
// If reached the end of the loop without exiting early,
// undo the final increment to avoid a null-pointer exception
if (i === path.length - 1) {
  i -= 1;
}
```
* a test for the broken function

### DIFF
--- a/modules/layers/src/layers/path-marker-layer/create-path-markers.ts
+++ b/modules/layers/src/layers/path-marker-layer/create-path-markers.ts
@@ -72,8 +72,8 @@ function createMarkerAlongPath({ path, percentage, lineLength, color, object, pr
   const distanceAlong = lineLength * percentage;
   let currentDistance = 0;
   let previousDistance = 0;
-
-  for (let i = 0; i < path.length - 1; i++) {
+  let i = 0;
+  for (i = 0; i < path.length - 1; i++) {
     currentDistance += path[i].distance(path[i + 1]);
     if (currentDistance > distanceAlong) {
       break;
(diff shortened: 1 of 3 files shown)


NameFILIP:
@supersonicclay @igorDykhta @ibgreen can you please prioritize to review/land/release this hotfix 🙏

I added a test for the affected function, but I think an even better solution would be to add visual regression tests to this repo.

modules/layers/src/layers/path-marker-layer/create-path-markers.ts
```diff
@@ -72,23 +72,26 @@ function createMarkerAlongPath({ path, percentage, lineLength, color, object, pr
...
     if (currentDistance > distanceAlong) {
       break;
     }
     previousDistance = currentDistance;
   }
 
-  const lastPoint = path[path.length - 1];
-  const beforeLastPoint = path[path.length - 2];
+  // If reached the end of the loop without exiting early,
+  // undo the final increment to avoid a null-pointer exception
```
NameFILIP, line 85: Why this is needed is better described here: https://github.com/uber/nebula.gl/issues/669

igorDykhta (approved):

NameFILIP:
@igorDykhta Thanks a lot for your quick actions!

PR #753, opened 2022-04-26
Fix issue where overlay items ignored z-index

austintang:
This diff fixes a bug introduced by a previous [diff](https://github.com/uber/nebula.gl/pull/747/files#diff-e1dc3f6f987fad6da6bee4154e7b2d5055535ead0d6f2120f90a21ee2b60b8a7R23) of mine where adding a z-index style to HTML overlay items wouldn't take effect.

The bug made it so that the following effect wasn't possible to create:
z-index

### DIFF
--- a/modules/overlays/src/html-overlay-item.tsx
+++ b/modules/overlays/src/html-overlay-item.tsx
@@ -14,13 +14,14 @@ type Props = {
 export default class HtmlOverlayItem extends React.Component<Props> {
   render() {
     const { x, y, children, style, coordinates, ...props } = this.props;
+    const { zIndex, ...remainingStyle } = style || {};
 
     return (
       // Using transform translate to position overlay items will result in a smooth zooming
       // effect, whereas using the top/left css properties will cause overlay items to
       // jiggle when zooming
-      <div style={{ transform: `translate(${x}px, ${y}px)` }}>
-        <div style={{ position: 'absolute', userSelect: 'none', ...style }} {...props}>
+      <div style={{ transform: `translate(${x}px, ${y}px)`, position: 'absolute', zIndex }}>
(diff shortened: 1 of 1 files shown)


igorDykhta (approved):

modules/overlays/src/html-overlay-item.tsx
```diff
@@ -14,13 +14,14 @@ type Props = {
...
+    const { zIndex, ...remainingStyle } = style || {};
 
     return (
       // Using transform translate to position overlay items will result in a smooth zooming
       // effect, whereas using the top/left css properties will cause overlay items to
       // jiggle when zooming
-      <div style={{ transform: `translate(${x}px, ${y}px)` }}>
-        <div style={{ position: 'absolute', userSelect: 'none', ...style }} {...props}>
+      <div style={{ transform: `translate(${x}px, ${y}px)`, position: 'absolute', zIndex }}>
+        <div style={{ userSelect: 'none', ...remainingStyle }} {...props}>
```
NameFILIP, line 24: what about the `userSelect: 'none'` should it be on this div or on the parent? 
in the [previous PR](https://github.com/uber/nebula.gl/pull/747) you introduced this nested div which made things more complicated and lead to this issue. Why is this nesting needed?

austintang: The nesting is needed because there are potentially two different transforms applied to each overlay item. The first being the transform the HTML overlay applies to each item to place them on the map, and the other potential transform would the one that the user applies to move the center/anchor point of each element, for example, from the default top left to bottom center for pins.

From what I can tell, `userSelect` can be on either level so I left it where it is. I moved the `position` style to the top level div because we need a `position` style for `zIndex` to work correctly.

NameFILIP (approved):

PR #772, opened 2022-06-27
Improving of feature geometry translation

IlyaIzr:
To translate features Nebula used turf `transfromTranslate` function for each point of feature. When applied to polygons this could've create a huge shape distortion
image
image
image

This approach helps to avoid such effect by first moving center point (with same function as turf/transformTranslate uses), and then building new geometry around new center point saving geometry proportions

Before:
image
After:
image

### DIFF
--- a/modules/edit-modes/test/translateFromCenter.test.ts
+++ b/modules/edit-modes/test/translateFromCenter.test.ts
@@ -1,0 +1,140 @@
+
+import type { Feature as TurfFeature, Geometry } from '@turf/helpers';
+// I used types from geojson because they're generic and helps me with intellisense
+import type {
+  Feature as GenericFeature,
+  Point,
+  MultiPoint,
+  LineString,
+  Polygon,
+  MultiPolygon,
+} from 'geojson';
+import { translateFromCenter } from '../src/translateFromCenter';
+
+type Feature = TurfFeature<Geometry>;
+
+test('Point coordinates in right format', () => {
+  const feature: GenericFeature<Point> = {
+    type: 'Feature',
+    geometry: {
+      type: 'Point',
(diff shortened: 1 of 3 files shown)


Akiyamka:
Looks like ti broke translate-mode.test.ts

modules/edit-modes/test/translateFromCenter.test.ts
```diff
@@ -0,0 +1,140 @@
...
+      coordinates: [0, 0],
+    },
+    properties: {},
+  };
+  const result = translateFromCenter(
+    feature as Feature,
+    100,
+    100,
+  ) as GenericFeature<Point>;
+  expect(result.geometry.coordinates).toHaveLength(2);
```
Akiyamka, line 30: Typescript already check interfaces, lets check transformation fix instead

IlyaIzr: It's kinda hard to test transformation because this approach doesn't defeat all the distortion - just makes it much smaller and visibly unnoticable, so the geometry wouldn't be strictly the same after moving feature and placing it right back

Akiyamka: Snapshot testing should work fine

IlyaIzr:
closing because similar https://github.com/uber/nebula.gl/pull/808 was already merged in

doneel:
I found this PR while searching google about the same underlying problem: using Turfjs`transformTranslate` over long distances causes distortion. While I'm working with a different mapping framework, this solution inspired a similar solution. Thank you @IlyaIzr for saving me a ton of headache, and hopefully others searching on this problem.

PR #774, opened 2022-06-28
Handle escape hotkey in linestring and polygon modes

### LINKED ISSUE #773: Escape hotkey doesn't work on linestring and polygon modes
While handleKeyUp has been implemented for the base [GeoJsonEditMode](https://github.com/uber/nebula.gl/blob/17aa19903bda8e5caaf14b6d25da624a1d317919/modules/edit-modes/src/lib/geojson-edit-mode.ts#L262), which calls resetClickSequence() when Escape is pressed, this is overridden in [DrawPolygonMode ](https://github.com/uber/nebula.gl/blob/17aa19903bda8e5caaf14b6d25da624a1d317919/modules/edit-modes/src/lib/draw-polygon-mode.ts#L124)and [DrawLineStringMode](https://github.com/uber/nebula.gl/blob/17aa19903bda8e5caaf14b6d25da624a1d317919/modules/edit-modes/src/lib/draw-line-string-mode.ts#L50)

The escape logic should presumably be moved to the polygon and linestring modes.  DrawPolygonByDraggingMode may need additional logic to cancel the dragging.

timnyborg:
Closes #773

Moves handling of keyup.escape to DrawLineStringMode, DrawPolygonMode, and DrawPolygonByDraggingMode.  In dragging mode, escape also halts the dragging handler, so no new points are added until a new click.

Also adds some hotkey tests.

### DIFF
--- a/modules/edit-modes/src/lib/geojson-edit-mode.ts
+++ b/modules/edit-modes/src/lib/geojson-edit-mode.ts
@@ -259,17 +259,7 @@ export class GeoJsonEditMode implements EditMode<FeatureCollection, GuideFeature
   handleStopDragging(event: StopDraggingEvent, props: ModeProps<FeatureCollection>): void {}
   handleDragging(event: DraggingEvent, props: ModeProps<FeatureCollection>): void {}
 
-  handleKeyUp(event: KeyboardEvent, props: ModeProps<FeatureCollection>): void {
-    if (event.key === 'Escape') {
-      this.resetClickSequence();
-      props.onEdit({
-        // Because the new drawing feature is dropped, so the data will keep as the same.
-        updatedData: props.data,
-        editType: 'cancelFeature',
-        editContext: {},
-      });
-    }
-  }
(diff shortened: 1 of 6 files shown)


modules/edit-modes/src/lib/geojson-edit-mode.ts
```diff
@@ -259,17 +259,7 @@ export class GeoJsonEditMode implements EditMode<FeatureCollection, GuideFeature
...
-      this.resetClickSequence();
-      props.onEdit({
-        // Because the new drawing feature is dropped, so the data will keep as the same.
-        updatedData: props.data,
-        editType: 'cancelFeature',
-        editContext: {},
-      });
-    }
-  }
+  handleKeyUp(event: KeyboardEvent, props: ModeProps<FeatureCollection>): void {}
```
igorDykhta, line 262: Restore this file. By removing this you are breaking Esc key event for many extended GeoJsonEdit modes.

igorDykhta: Never mind, I'll do it.

igorDykhta (approved):

PR #775, opened 2022-07-05
fix isFeatureSelected performance

timnyborg:
`isFeatureSelected()` becomes a performance bottleneck when editing layers with 30,000+ features, because it's running `indexOf` on every single feature (so O(n^2) time, regardless of how many features are selected).

With this change, each feature is checked against the contents of the selectedFeatures list, which is O(n) if you have a single item selected, gradually taking more time as the list increases.  This could probably be improved further by using a Set for selectedFeatures, but I don't know the implications of altering the structure of the layer's `state`.


modules/layers/src/layers/editable-geojson-layer.ts
```diff
@@ -444,8 +444,7 @@ export default class EditableGeoJsonLayer extends EditableLayer<
     if (!this.props.selectedFeatureIndexes.length) {
       return false;
     }
-    const featureIndex = this.props.data.features.indexOf(feature);
-    return this.props.selectedFeatureIndexes.includes(featureIndex);
+    return this.state.selectedFeatures.some(selected => selected === feature)
```
igorDykhta, line 447: return this.state.selectedFeatures.some(feature)

timnyborg: I'm guessing you meant `.includes(feature)`

Akiyamka (approved):
Can confirm, performance improved. Nothing seems broken

igorDykhta (approved):

PR #777, opened 2022-07-18
correct typing on EditableGeojsonLayerProps get* functions

timnyborg:
The typing of getLineWidth, getRadius, etc. doesn't reflect that they're passed isSelected and mode, like all the others


igorDykhta (approved):

modules/layers/src/layers/editable-geojson-layer.ts
```diff
@@ -133,12 +133,12 @@ export type EditableGeojsonLayerProps<DataT = any> = EditableLayerProps<DataT> &
...
   getFillColor?: Color | ((feature, isSelected, mode) => Color);
-  getRadius?: number | ((f) => number);
-  getLineWidth?: number | ((f) => number);
+  getRadius?: number | ((feature, isSelected, mode) => number);
+  getLineWidth?: number | ((feature, isSelected, mode) => number);
 
   getTentativeLineColor?: Color | ((feature, isSelected, mode) => Color);
   getTentativeFillColor?: Color | ((feature, isSelected, mode) => Color);
-  getTentativeLineWidth?: number | ((f) => number);
+  getTentativeLineWidth?: number | ((feature, isSelected, mode) => number);
```
Akiyamka, line 141: For me it looks like type was correct.

There how it used:
https://github.com/uber/nebula.gl/blob/master/modules/layers/src/layers/editable-geojson-layer.ts#L203
https://github.com/uber/nebula.gl/blob/8697d51528b8fcbc9b3dbb6482cdd7e383cebce8/modules/layers/src/layers/editable-geojson-layer.ts#L72

Am I missing something?

igorDykhta: You are correct. This one uses different accessor.

timnyborg: Yep, will fix

PR #828, opened 2022-11-28
fix: Fix to fire the finishMovePosition state at the end of a feature…

giuseppe-straziota:
During the editing mode, the state called finishMovePosition is never fired, I found out that the problem was derived by a mistyped call to an object property and to a comparison between strings, in particular with the constant ELEMENT_TYPE.FEATURE that in my case has a capital letter differently from what is defined in the code, for this reason, I added a call to toLowerCase function to the switch because of the many places where the constant is used.

### DIFF
--- a/modules/react-map-gl-draw/src/edit-modes/editing-mode.ts
+++ b/modules/react-map-gl-draw/src/edit-modes/editing-mode.ts
@@ -73,12 +73,12 @@ export default class EditingMode extends BaseMode {
     const picked = event.picks && event.picks[0];
 
     // @ts-ignore
-    if (!picked || !picked.Object || !isNumeric(picked.featureIndex)) {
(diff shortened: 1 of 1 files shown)


igorDykhta (approved):
Nice

igorDykhta:
@giuseppe-straziota Hi, could you sign Contributor License Agreement?

giuseppe-straziota:
Hi @igorDykhta happy to be useful to the project. I signed but seems it's still pending.

igorDykhta:
@giuseppe-straziota you should sign with the correct email/account. 

> giuseppestraziota seems not to be a GitHub user. You need a GitHub account to be able to sign the CLA. If you have already a GitHub account, please [add the email address used for this commit to your account](https://help.github.com/articles/why-are-my-commits-linked-to-the-wrong-user/#commits-are-not-linked-to-any-user).

PR #835, opened 2022-12-05
feat: allow billboard in points-circle guides layer

Radik24:
Allows billboard for points-circle guides layer


igorDykhta (commented):
Do we need a separate billboard prop for guide layers?
Pass billboard prop for points-icon guide layer as well.

Radik24:
Hi

@igorDykhta I often do measuring on 3D data and it looks much better. If you wouldn't mind I would like to have this option. I'll update `points-icon`.

PR #873, opened 2023-03-09
Correct onEdit typing

timnyborg:
Brings it into line with how it's actually used: https://github.com/uber/nebula.gl/blob/a3f81bb6476b36a34a20359c97046c3a75b3ecdd/modules/edit-modes/src/types.ts#L129
https://github.com/uber/nebula.gl/blob/2a962399ef7c45c1bda8ddb34394becb04a08995/modules/layers/src/layers/editable-geojson-layer.ts#L413


igorDykhta (approved):

igorDykhta:
@timnyborg try to update editable-h3-cluster-layer as well

timnyborg:
> @timnyborg try to update editable-h3-cluster-layer as well

Done

PR #900, opened 2024-04-11
feat: polygon prevent overlapping in ModifyMode

anton-gustafsson:
Hi! this PR adds preventOverlappingLines support for ModifyMode.

Also fixed an edge case in drawPolygon if the first dot was pressed (closing the polygon) the line could still be overlapped

### DIFF
--- a/modules/edit-modes/src/lib/draw-polygon-mode.ts
+++ b/modules/edit-modes/src/lib/draw-polygon-mode.ts
@@ -8,8 +8,8 @@ import {
   TentativeFeature,
   GuideFeature,
 } from '../types';
-import { Polygon, FeatureCollection } from '../geojson-types';
-import { getPickedEditHandle } from '../utils';
+import { Polygon, FeatureCollection, Position } from '../geojson-types';
+import { getPickedEditHandle, hasPolygonCrossingLines } from '../utils';
 import { GeoJsonEditMode } from './geojson-edit-mode';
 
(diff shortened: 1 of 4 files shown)


modules/edit-modes/src/lib/draw-polygon-mode.ts
```diff
@@ -119,9 +119,21 @@ export class DrawPolygonMode extends GeoJsonEditMode {
         coordinates: [[...clickSequence, clickSequence[0]]],
       };
 
+      const editAction = this.getAddFeatureOrBooleanPolygonAction(polygonToAdd, props);
+
+      if (
```
anton-gustafsson, line 124: This fixes If first editHandle was clicked and last line being created would overlapp

modules/edit-modes/src/lib/draw-polygon-mode.ts
```diff
@@ -119,9 +119,21 @@ export class DrawPolygonMode extends GeoJsonEditMode {
...
+        props.modeConfig?.preventOverlappingLines &&
+        editAction.updatedData &&
+        hasPolygonCrossingLines(
+          editAction.updatedData.features[editAction.editContext.featureIndexes[0]].geometry
+            .coordinates[0] as Position[]
```
anton-gustafsson, line 129: Any ideas to remove this cast to Position[]?

